PR #1, opened 2015-09-02
[ADD] membership_variable_period

pedrobaeza:
# Variable period for memberships

Current membership module allows to set products that define a fixed period
for a membership. This is good when the quotas are defined periodically, and
when you become a member, you are until the end of that quota cycle.

But a lot of times, membership quotas express an amount of time that you
gain the membership. For example, one year since the subscription.

This module allows to make it in Odoo, using current membership features,
and adapting them for this purpose. As now the quota is not attached to a fixed
period, you can also invoice more than one quantity for being a member for
the corresponding number of periods.

Finally, a cron has been included that triggers the recalculation of the
membership state, allowing to have "old members", which doesn't work well
on standard.
# Usage

Define a member product, and select 'Variable periods' in the field
_Membership type_. You will be able to select them the period for the
membership in quantity and interval.


pedrobaeza:
@rafaelbn @antespi You can start reviewing it

rafaelbn:
Hi @pedrobaeza I cannot select variable period. I have checked the module is installed. See screenshot:

pedrobaeza:
@rafaelbn, @antespi, problems solved and tests added.

janetrp:
works fine, very useful as association products :+1:

rafaelbn:
Again tested in runbot :+1:

antespi:
Why start is 1st Octuber? when product is 1 week and buy membership at 4th September?

Product:
screenshot from 2015-09-04 15 10 34

Customer form, membership tab:
screenshot from 2015-09-04 15 11 12

NOTE: this product was saved as fix dates at first time, then modified to variable with 1 week

janetrp:
@antespi have you buyed an association before? It seems same partner can´t have several associations.
I've just checked you have 2 invoices for the same partner
captura2

membership_variable_period/models/product_template.py
```diff
@@ -0,0 +1,63 @@
...
+    membership_interval_qty = fields.Integer(
+        string="Interval quantity", default=1)
+    membership_interval_unit = fields.Selection(
+        selection=[('days', 'days'),
+                   ('weeks', 'weeks'),
+                   ('months', 'months'),
+                   ('years', 'years')],
+        string="Interval unit", default='years')
+
+    def _check_vals_membership_type(self, vals, membership_type):
```
antespi: I propuse to implement this like:

```
def _check_vals_membership_type(self, vals, membership_type):
   if membership_type == 'variable':
      vals['membership_date_from'] = False
      vals['membership_date_to'] = False
   return vals
```

Because in any case, when `membership_type == 'variable'`, dates field must be False

pedrobaeza:
This one is also ready to be merged.

antespi:
:+1: tested again in runbot and it works. Thanks

rafaelbn:
Againtest tested. :+1:

PR #2, opened 2015-09-02
[ADD] membership_initial_fee

pedrobaeza:
# Initial fee for memberships

Charge an initial fee when a partner is invoiced for the first time with a
member product.
# Usage

Define a member product, and select 'Fixed amount' or 'Percentage of the price'
for invoicing an extra charge in the first invoice that is created with this
member product.


pedrobaeza:
Another one to review @rafaelbn @antespi

rafaelbn:
Hi Pedro, just one doubt:
- Initial fee is added to the invoice withour a product. How can I analyze each year what revenue I have with initial fees?

pedrobaeza:
@rafaelbn, added product to the initial fee line

janetrp:
with the Pedro improvement we should create first the product membership without the initial fee and add the other product making reference to the product we want to apply the initial fee. 
It works fine :+1:

rafaelbn:
@pedrobaeza 
- When creating the product for "Product for initial fee" from Association -> Membership Products it is created as Consumable and must be created as Service
- When selecting a product in "Product for initial fee" doesn't bring the sale price from the product to the field "Fixed fee"

membership_initial_fee/models/product_template.py
```diff
@@ -0,0 +1,19 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    initial_fee = fields.Selection(
+        selection=[('none', 'No initial fee'),
+                   ('fixed', 'Fixed amount'),
+                   ('percentage', 'Percentage of the price')],
+        default='none', string="Initial fee", required=True)
+    fixed_fee = fields.Float(digits_compute=dp.get_precision('Product Price'))
```
antespi, line 16: IMHO `fixed_fee` should be:

```
fixed_fee = fields.Float(digits_compute=dp.get_precision('Product Price'), related='product_fee.list_price', readonly=True)
```

Because now there are two price fields to configure, one in fee product (fixed_fee) and another in initial fee product (list_price)  and this could be consfusing for user.

And shoud have a new boolean field to indicate that product is an initial fee

```
is_initial_fee = fields.Boolean()
```

And configure this domains:
- Show `membership = True` and `is_initial_fee = False` in Membership Products list
- Show `membership = True` and `is_initial_fee = False` in 'Membership' one2many field at wizard when buying a new membership
- Show `membership = True` and `is_initial_fee = True` in 'Product for initial fee' one2many field in membership product form view

By default, when creating a new initial fee product from 'Product for initial fee' one2many field, should have `membership = True` and `is_initial_fee = True` by default

Then we can see all membership related products (fees and initial fees) in 'Members Analysis' report

In case, 'initial_fee' = percentage, IMO current implementation is OK.

antespi:
When buying a membership product with a initial fee by percent, in invoice line description is not showing percent applied. It could be useful for customer to add applied percent in invoice line description.

membership_initial_fee/models/account_invoice.py
```diff
@@ -0,0 +1,65 @@
...
+            product_fee.id, False, line_vals['quantity'], '',
+            'out_invoice', invoice_line.invoice_id.partner_id.id,
+            invoice_line.invoice_id.fiscal_position.id)
+        line_vals.update(line_dict['value'])
+        if product.initial_fee == 'fixed':
+            line_vals['price_unit'] = product.fixed_fee
+        elif product.initial_fee == 'percentage':
+            line_vals['price_unit'] = (
+                product.percentage_fee * invoice_line.price_unit / 100)
+        line_vals['name'] = _('Membership initial fee')
```
antespi: Add percent applied in 'name' field if ```product.initial_fee == 'percentage'```

antespi:
:+1: I realize that these changes will implied more issues in number of membership_line (generates two) and in membership analysis

JavierIniesta:
Tested on runbot.

 :+1:

pedrobaeza:
Changes requested by @rafaelbn done

rafaelbn:
Again tested in runbot :+1:

PR #3, opened 2015-09-02
[ADD] membership_prorrate

pedrobaeza:
# 
# Prorrate the membership fee

# Usage

Go to the membership product, and check "Prorrate" in it. When you create
invoices that includes a membership product by hand or pressing the button
"Buy Membership" from members form.


pedrobaeza:
cc @rafaelbn @antespi

rafaelbn:
Tested in runbot :+1:

membership_prorrate/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': 'Prorrate membership fee',
+    'version': '8.0.1.0.0',
+    'category': 'Association',
```
antespi, line 9: no license key, please add

```
'license': 'AGPL-3',
```

janetrp:
it works :+1:

JavierIniesta:
Tested on runbot.

:+1:

membership_prorrate/models/account_invoice.py
```diff
@@ -0,0 +1,57 @@
...
+        date_to = fields.Date.from_string(product.membership_date_to)
+        return date_from, date_to
+
+    def _prepare_invoice_line_prorrate_vals(self, invoice_line):
+        product = invoice_line.product_id
+        date_invoice = fields.Date.from_string(
+            invoice_line.invoice_id.date_invoice or fields.Date.today())
+        date_from, date_to = self._get_membership_interval(
+            product, date_invoice)
+        if date_invoice < date_from:
```
antespi: If you combine this addon with this one: https://github.com/OCA/vertical-association/pull/1 and product is variable (not fixes dates), then `date_from` and `date_to` will be False. So a python error like this will occur:

```
TypeError: can't compare datetime.date to bool
```

So I suggest to add a condition to avoid this case.

antespi:
:+1:, after fixes commented. Thanks for this addon, it's very useful.

rafaelbn:
Again tested in runbot :+1:

membership_prorrate/models/account_invoice.py
pedrobaeza: That's because there's one auto-installed module to join both that I will publish next, that overwrites `_get_membership_interval` to handle the situation with prorrate, so no problem in this one.

pedrobaeza:
@antespi, license added

pedrobaeza:
This one is also ready and at 100% coverage

antespi:
:+1: thanks

rafaelbn:
If you have a membership which cost 120€ from 01/01/2015 to 31/12/2015 and you make a member today, invoice says amount to pay is 38.52 but membership says that has pay 120.00 and that membership started 01/01/2015 and not today 05/09/2015. I think It should say correct dates and amount.

captura
captura2

pedrobaeza:
@rafaelbn, fixed and tested the price on membership line

rafaelbn:
Hi @pedrobaeza the Membership start date also must be fixed as amount.

pedrobaeza:
Fixed the membership start date

rafaelbn:
OK thanks! :+1:

PR #5, opened 2015-09-18
[FIX] membership_variable_period: Take into account invoice_id on write method

pedrobaeza:
A strange operation creating invoices from sales orders happens: invoice lines
are created without invoice_id, and later this field is assigned. This fix
takes this into account to fill membership date start and end properly
on the membership lines for this case.


rafaelbn:
Reproduced bug and fixed with this PR. Tested :+1:

dgzurita:
Tested :+1:

JavierIniesta:
Tested on runbot :+1:

pedrobaeza:
I have added the test case to keep coverage. If green, @rafaelbn please merge.

PR #7, opened 2015-09-22
[FIX] Optimizing check_membership_expiry partners search

antespi:
Optimizing partners search which are going to be check for membership expiration.
- **Free**: A partner marked as free must not to be checked
- **Associate member = False**: Only partners with no associate member must be check, because his children (partner with him as associate member) will be update automatically.

Please @pedrobaeza, check if this is conceptually right and could be applied for any installation, thanks.


pedrobaeza:
:+1:

JavierIniesta:
:+1:

dgzurita:
:+1:

PR #9, opened 2016-01-13
[FIX] Exclude current line when searching for other membership_lines

antespi:
This exclusion is important to ignore the currently created membership_line.

Error case:
- Create a contract starting at 01/01/2016 for a member partner, with a product like:
  - membership_type = 'variable'
  - membership_interval_qty = 1
  - membership_interval_unit = 'years'
  - membership_prorrate = True
- Create an invoice

Current behavior:
- Membership line created is not starting at 01/01/2016

Expected behavior:
- Membership line created should start at 01/01/2016

### DIFF
--- a/membership_variable_period/models/product_product.py
+++ b/membership_variable_period/models/product_product.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, api
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.multi
+    def _get_next_date(self, date, qty=1):
+        self.ensure_one()
+        return self.product_tmpl_id._get_next_date(date, qty=qty)
--- a/membership_prorrate_variable_period/models/__init__.py
+++ b/membership_prorrate_variable_period/models/__init__.py
@@ -3,3 +3,4 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from . import account_invoice
+from . import product_template
--- a/membership_prorrate_variable_period/models/account_invoice.py
+++ b/membership_prorrate_variable_period/models/account_invoice.py
@@ -21,24 +21,15 @@ class AccountInvoiceLine(models.Model):
         if product.membership_type == 'fixed':
             return super(AccountInvoiceLine, self)._get_membership_interval(
                 product, date)
-        if product.membership_interval_qty != 1:
-            raise exceptions.Warning(
-                _("It's not possible to prorrate periods which interval "
-                  "quantity is different from 1."))
         if product.membership_interval_unit == 'days':
             raise exceptions.Warning(
                 _("It's not possible to prorrate daily periods."))
         if product.membership_interval_unit == 'weeks':
             weekday = date.weekday()
             date_from = date - datetime.timedelta(weekday)
-            date_to = date_from + datetime.timedelta(6)
         elif product.membership_interval_unit == 'months':
             date_from = datetime.date(day=1, month=date.month, year=date.year)
-            last_month_day = calendar.monthrange(
-                date.year, date.month)[1]
-            date_to = datetime.date(
-                day=last_month_day, month=date.month, year=date.year)
         elif product.membership_interval_unit == 'years':
             date_from = datetime.date(day=1, month=1, year=date.year)
-            date_to = datetime.date(day=31, month=12, year=date.year)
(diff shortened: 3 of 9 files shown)


rafaelbn:
Travis fails

JavierIniesta:
Tested in runbot :+1:

antespi:
@JavierIniesta, please review again I've refactored membership_line changes propagation.

JavierIniesta:
Runbot and Travis fail.

pedrobaeza:
Is it required to change too much code and the signature of the methods? I think this is causing the tests in depending modules to fail.

antespi:
@pedrobaeza, yes you're right. I'm waiting to you for discussing this online, because I have several questions about the implementation. Resuming:
1. Current implementation if failing when creating an invoice with qty > 1. Test added in this PR shows that issue. This PR fix this use case.
2. Current implementation if failing when creating an invoice from a contract. This issue can be reproduce in runbot. This PR fix this use case.
3. This PR is generating fails in related modules and we have to fix them before merge.
4. In order to fix related modules test, could be possible to create only one membership line per invoice line? This change will simplify so much the code, but I don't know if there's a good reason for generating several membership_lines when invoice line qty is greater than 1.

Thanks.

antespi:
Just for remembering:
- Add a new field 'membership_start_force' to account.invoice model in order to force membership start date in related membership.membership_line
- This field is automatically set in create (if not defined) using invoice date or today
- Refactor relating only one membership.membership_line per account.invoice.line, calculating date_to  depending on product configuration and qty in invoice line.
- Integrate https://github.com/OCA/vertical-association/issues/10
- Fix travis red status introduced here https://github.com/odoo/odoo/commit/f6d2a75c94e9f3b6d64efefb1d5b7f599a10d576

pedrobaeza:
Second line is not correct. We don't set on create, but only test if present, and use it in that case. If not, we take the invoice date or the current date if also not present.

antespi:
Finally Travis is green :smile:

rafaelbn (approved):
Please @antespi @pedrobaeza last review before merging 👍 @Tecnativa

antespi:
I think this is ready to be merged

rafaelbn:
Please @antespi squash commits before mergin

membership_variable_period/models/product_product.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, api
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.multi
+    def _get_next_date(self, date, qty=1):
```
pedrobaeza, line 12: This is not needed, as product.product inherits all from product template.

antespi: Well, this may apply for fields but not for methods.
I've tested removing the method and tests crash with this error:

```
2016-11-17 17:11:02,407 15150 ERROR variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: ERROR: test_get_next_date (openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period.TestMembershipProrrateVariablePeriod)
2016-11-17 17:11:02,407 15150 ERROR variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: Traceback (most recent call last):
2016-11-17 17:11:02,407 15150 ERROR variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: `   File "/opt/odoo/v8/other-addons/membership_prorrate_variable_period/tests/test_membership_prorrate_variable_period.py", line 115, in test_get_next_date
2016-11-17 17:11:02,407 15150 ERROR variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: `     self.product._get_next_date(date))
2016-11-17 17:11:02,407 15150 ERROR variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: ` AttributeError: 'product.product' object has no attribute '_get_next_date'
2016-11-17 17:11:02,407 15150 INFO variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: Ran 7 tests in 2.021s
2016-11-17 17:11:02,407 15150 ERROR variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period: FAILED
2016-11-17 17:11:02,407 15150 INFO variable_period openerp.addons.membership_prorrate_variable_period.tests.test_membership_prorrate_variable_period:  (errors=5)

```

pedrobaeza: OK, it's the inheritance chain order problem that v10 solves. We have to stay as is now. Merging...

PR #10, opened 2016-01-13
[FIX] Use invoice date as join date if defined

antespi:
Join date (membership.membership_line.date) is forced to be today, but when creating an invoice from a contract, invoice date is setting at create time, so this improvement allows to set this date according to invoice.


rafaelbn:
Travis fails

antespi:

pedrobaeza:
Closed as discussed to be integrated in #9

PR #11, opened 2016-01-13
[FIX] Use the same analytic_account as normal membership invoice line

antespi:
When using contract (account.analytic.account) membership invoice line is created with analytic_account.

This improvement allows that inital_fee invoice line have the same analytic_account as membership invoice line by default.


rafaelbn:
Runbot and travis fails

antespi:

Runbot was killed, no error reported

pedrobaeza:
Error is unrelated, so I  fast-track. Solution will come in #9

PR #12, opened 2016-01-13
[IMP] See if partner has any membership line instead of and invoice

antespi:
Current behavior:
- To decide whether or not to create the initial fee, we search for a draft, open or paid invoice for that partner with the same membership product.

PR behavior:
- See if partner has any membership line to decide whether or not to create the initial fee, no matter what membership product or membership dates.
- Method is isolated in order to be inheritable for customization.

### DIFF
--- a/membership_initial_fee/models/account_invoice.py
+++ b/membership_initial_fee/models/account_invoice.py
@@ -46,7 +46,7 @@ class AccountInvoiceLine(models.Model):
             ('account_invoice_line', 'not in', (invoice_line.id, )),
             ('state', 'not in', ['none', 'canceled']),
         ])
-        return True if member_lines else False
+        return not bool(member_lines)
 
     @api.model
     def create(self, vals):
--- a/membership_initial_fee/models/account_invoice.py
+++ b/membership_initial_fee/models/account_invoice.py
@@ -33,18 +33,27 @@ class AccountInvoiceLine(models.Model):
         return line_vals
 
     @api.model
+    def initial_fee_create_check(self, product, invoice_line):
+        """Inherit this method to implement a custom method
+           to decide whether or not to create the initial fee
+        """
+        if not product.membership or product.initial_fee == 'none':
+            return False
+        # See if partner has any membership line to decide whether or not
+        # to create the initial fee
+        member_lines = self.env['membership.membership_line'].search([
+            ('partner', '=', invoice_line.invoice_id.partner_id.id),
+            ('account_invoice_line', 'not in', (invoice_line.id, )),
+            ('state', 'not in', ['none', 'canceled']),
+        ])
+        return True if member_lines else False
+
+    @api.model
     def create(self, vals):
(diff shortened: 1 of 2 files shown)


rafaelbn:
Travis fails

antespi:

Runbot was killed, no error reported

membership_initial_fee/models/account_invoice.py
```diff
@@ -32,18 +32,28 @@ def _prepare_initial_fee_vals(self, invoice_line):
...
+        if not member_lines:
+            # Charge initial fee
+            return self.create(self._prepare_initial_fee_vals(invoice_line))
+        return False
+
+    @api.model
     def create(self, vals):
         invoice_line = super(AccountInvoiceLine, self).create(vals)
         product = self.env['product.product'].browse(
             vals.get('product_id'))
```
mourad-ehm, line 55: product_id is not required on invoice line. to avoid a bug on line without product_id, It will be nice to test if product_id exist before pass to the rest of code.

pedrobaeza, line 55: This is not a problem, because `browse(None)` returns an empty recordset, and checks for this empty recordset are consistent.

pedrobaeza:
Please refactor this PR to have the method as a test returning False or True instead of creating the line inside the method.

Add also in the known issues the features we can add at product level for discerning if we should add the initial fee.

antespi:
@pedrobaeza fixes done. Thanks

membership_initial_fee/models/account_invoice.py
```diff
@@ -33,18 +33,27 @@ def _prepare_initial_fee_vals(self, invoice_line):
...
+        if not product.membership or product.initial_fee == 'none':
+            return False
+        # See if partner has any membership line to decide whether or not
+        # to create the initial fee
+        member_lines = self.env['membership.membership_line'].search([
+            ('partner', '=', invoice_line.invoice_id.partner_id.id),
+            ('account_invoice_line', 'not in', (invoice_line.id, )),
+            ('state', 'not in', ['none', 'canceled']),
+        ])
+        return True if member_lines else False
```
pedrobaeza: `return bool(member_lines)`

antespi:
@pedrobaeza, return fixed, thanks.

pedrobaeza:
:+1: Travis will be fixed in the other PR.

@rafaelbn @JavierIniesta, please review

pedrobaeza:
@antespi, is Travis still failing on main branch?

rafaelbn (approved):
Thanks

pedrobaeza:
@rafaelbn, my question was not answered, and the implementation is not correct, so you shouldn't merge this. @antespi, please propose another one with the correct implementation.

rafaelbn:
It was answered here https://github.com/OCA/vertical-association/pull/12#issuecomment-171456821

pedrobaeza:
No, but anyway, you did it. We'll fix it later.

PR #13, opened 2016-02-14
[8.0][ADD] membership_extra_info addon

antespi:
This module extends the functionality of membership to add extra useful
membership information

A partner could be member in one period (from Jan 2001 to Oct 2003) and then
signup again two years later (from Feb 2005). Standard Odoo membership addon
considers membership start date as Jan 2001, instead of Feb 2005.

Also a partner could request us for decline his membership (Ago 2006) but
desire to end his membership later (Sep 2006) and we can classify decline
reasons for later analysis

An finally we can define different membership categories, for instance,
Bronce, Silver and Gold. Membership product could be related with a membership
category, so partner membership category is the last membership line product category
- Membership last start date: Start date of last membership period, Feb 2005 in our example
- Membership decline date: Decline date of current membership period, Ago 2006 in our example
- Membership decline reason: Decline reason of current membership period, from an admin defined list of reasons
- Membership category: Membership category of current membership period, from an admin defined list of categories
# Configuration

You can define decline reasons at Association > Configuration > Membership decline reasons,
to be selected when a partner request us to decline his membership

You can define membership categories at Association > Configuration > Membership categories,
to be selected in membership product form
# Usage

When a partner has a valid (paid) membership line a new button (X) appears. Click
on it when partner requests us to decline his membership. Then a pop-up appears to
define:
- Decline date: Request date, normally today
- New finish date: Finish date of current membership period requested by partner
- Decline reason

After that you can create a refund if necessary, clicking at membership line
invoice and clicking at 'Ask Refund' button.


pedrobaeza:
At first sight, I will split this module in 2: one for the extra info (only display), and another for the extra categories.

antespi:
Extra categories are for info and classification purposes, that's why there's only one addon. No new flow or logic added here.

Travis errors related with `TestMembershipProrrateVariablePeriod`. Those errors will be fixed in other PR

pedrobaeza:
OK, I understand. It's good for me.

max3903:
@antespi I am not sure to understand the feature of this module:
- Does it allow any date as a start and end date? Looks very much like membership_variable_period
- Does it provide the ability to terminate the membership earlier and specify a reason? In this case, I would rename it to membership_early_termination and provide a setting to configure the reason as mandatory or not.

I try to avoid "extra", "custom", "improvement" or "enhance" in the module name. Many module could claim the same name with those terms, but there is one only way that your module improves membership. So try to explain that in the name. You will get better visibility by being well referenced.

andhit-r:
Agree with @max3903.

rafaelbn:
Hi @max3903 @andhit-r I will explain it:

"@antespi I am not sure to understand the feature of this module:"

"Does it allow any date as a start and end date? Looks very much like membership_variable_period"
- This is absolutely different concept.

"Does it provide the ability to terminate the membership earlier and specify a reason? In this case, I would rename it to membership_early_termination and provide a setting to configure the reason as mandatory or not."
- Much more functionality

Let's go, Odoo core you have 2 fields:
- Membership start date (date from which you are member in the **last invoce**)
- Membership end date (date until you are partners in the **last invoice**)

Then you need more data:
- Membership historical start date, in a Gym for ex. you are member from 2016/06/28 but actual membership date is 2016/02/01
- Membership decline date, in a Gym you have paid membership until 2016/02/29, if you want to cancel you membership today and not wait until end of month you need this fields. Remember this case when you give to opportunity to you customer choosing end date (today / until memberships ends)
- No doubt about cancel reason (isn't it?)
- Category, Odoo make relation the product with category but in the cases you have different products for the same category this doesn't work. Ex: a Gold partner is that company who pay different amount depending of his capital, employees or other criteria. then different companies who pay different products are the same category.

I think problem here is module's name: membership_extra_info addon and README? What name do you think is better @max3903 @pedrobaeza @andhit-r ?

Said that, @antespi it looks dates are not working properly. Check:

max3903:
@antespi Something like `membership_actual_date` or `membership_date_not_from_invoice`

rafaelbn:
Hi, my comment https://github.com/OCA/vertical-association/pull/13#issuecomment-184684728 is an user explanation. But I mixed Odoo core functionality, other OCA modules functionality. So I do not explain correctly this PR.

All agree we should change name and @antespi will clarify.

rafaelbn (approved):
Please @antespi review comments and check travis, this module is ready to merge after comments cc @pedrobaeza @Tecnativa

antespi:
I close this PR in favor to https://github.com/OCA/vertical-association/pull/21 nd https://github.com/OCA/vertical-association/pull/22

PR #15, opened 2016-04-09
[ADD] membership_autoextend

hbrunn:
# 
# Autoextend membership

This module was written to define membership products as being autoextended.

Members with a membership of this type will be informed before an extension takes place in order to give them the chance to cancel their membership.
# Configuration

To configure this module, you need to:
- go to a membership product
- check box `Autoextend membership`
- optionally set email templates to inform a member about an upcoming extension and after it's done
- optionally set an amount of days before the extension the warning mail should be sent
- if you select an extension membership, this product will be used to extend the membership. Otherwise, the current membership product will be used


hbrunn:
(I took the liberty to fix the test for `membership_prorrate_variable_period` in https://github.com/OCA/vertical-association/pull/15/commits/a0d7ae05449fb75be8a696037bf1ef379819c1e6)

pedrobaeza:
@hbrunn, thanks for the module and the patch of the tests. This module is another approach for the same purpose of _membership_variable_period_ module: have an auto-renewable product. And I see better the approach of the current module, that put directly the period of the membership, allowing the combination with the rest of the modules. Please check it.

hbrunn:
aren't they complementary? Your module gives us the variable periods, that's important for my project too, but with this addon, you can have memberships that are renewed without user interaction. With _membership_variable_period_, somebody still has to do manual actions to renew a membership, right?

pedrobaeza:
I'm using for that the contracts with recurring invoicing with this product.

pedrobaeza:
With that, you have the new invoices generated (which are important and the control you have over renewals), and when paid, the subscription is renewed. Maybe I should write that on the README.

hbrunn:
That's certainly a good idea. Can't we have both? In my case, stuff needs to be as uncomplicated as possible, I can imagine this is useful for other people too, especially the notification part. You could do this with automated actions too, but that's not as convenient as setting all the stuff on the product.

pedrobaeza:
OK, let's clarify of all this in the READMEs and we can keep both.

leemannd:
@pedrobaeza @hbrunn IMHO It is ready to be merged.
:+1:

pedrobaeza:
@hbrunn, please add the clarification we talked about on the README and I merge.

hbrunn:
@pedrobaeza done

membership_autoextend/README.rst
```diff
@@ -20,6 +20,8 @@ To configure this module, you need to:
 * optionally set an amount of days before the extension the warning mail should be sent
 * if you select an extension membership, this product will be used to extend the membership. Otherwise, the current membership product will be used
 
+Note that this module is not strictly necessary for automatic extensions, your can have the same effect with the modules ``subscription`` and ``membership_variable_period``.
```
pedrobaeza: s/your/you

hbrunn: thanks, fixed

pedrobaeza (approved):

pedrobaeza:
Strange, Travis OCB is now not happy with your test fix...

PR #16, opened 2016-04-27
[ADD] website_membership_signup

hbrunn:
# 
# Sign up via the website

This module was written to allow website visitors to sign up for the
association. It does _not_ handle online payments. For this, a glue module
for the `payment` module would be necessary, which doesn't exist currently.
# Configuration

To configure this module, you need to:
- activate the checkbox `website_published` for at least one membership
  product
- activate the page `Member sign up`, which is activated by default after
  installing this module
# Usage
- Users sign up via the website
- They have to choose a membership product
- They will be invoiced with this product and end up as invoiced members
- As soon as the invoice is paid (usually via a direct debit), the person
  becomes a paid member


rafaelbn:
Thanks!
- Please check travis.
- Filter by membership in form:
  
- I recommend (IMHO is mandatory for security) to verify email. You can use: https://github.com/OCA/server-tools/tree/8.0/auth_signup_verify_email
- I have tested and you create directly a **validated invoice** without checking any field of the form. I think:
  - You should check at least Account number and BIC
  - Why not to add VAT number field?
  - The invoice should be create in draft for reviewing before validate
    This is the result of a dummy form:

andhit-r:
@hbrunn thanks for submitting this great module. I've tested on runbot and bellow are my thoughts:
- If this module intend to be _sign up_ form for membership, i think there should be an automatic mechanism to create user. If no automatic user creation will provide along with this module, then i would suggest that you put it on Known Issue so whoever implement this module create their own mechanism.
- I agree with @rafaelbn  regarding created invoice. It should be on draft state.

hbrunn:
@rafaelbn travis is fixed in #17, thanks for the point with the filter. Verifying email addresses is out of the scope of this module, as is @andhit-r's user creation.
That the invoice is validated is what my customer wants, and VAT is not relevant for a lot of noncommercial associations. Whoever wants/needs that is cordially invited to add the appropriate configuration options in subsequent PRs.
Account numbers are checked if the module base_iban is installed.

hbrunn:
tests missing

PR #21, opened 2016-11-18
[8.0][ADD] membership_extension

antespi:
Membership extension
====================

This module extends Odoo's membership management with the following features:

* Membership Category
* Membership lines editable
* Do not calculate membership state from invoice status
* Start date of last membership period

Configuration
=============

Users can define membership categories in Association > Configuration > Membership Categories
Then go to membership products and set a category to each one

Usage
=====

Membership lines are created when a membership product is invoiced, like in
Odoo standard version. But now users can create a new membership line without
creating an invoice.

Also, users can edit membership line dates and state even an invoice is related
with it.


antespi:
This error is related with this: https://github.com/odoo/odoo/issues/3922#issuecomment-129874599
For testing in runbot, use base database

membership_autoextend/tests/test_membership_autoextend.py
```diff
@@ -57,10 +57,6 @@ def invoice_membership(self, partner, membership_product, extra_days=0):
             'date_invoice': fields.Date.to_string(invoice_date),
         })
         invoice.signal_workflow('invoice_open')
-        # trigger membership_state update
```
pedrobaeza, line 60: Why do you remove this?

membership_extension/README.rst
```diff
@@ -0,0 +1,68 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Membership extension
+====================
+
+* Membership Category
```
pedrobaeza: Include a preface saying: `This module extends Odoo's membership management with the following features:`

membership_extension/README.rst
```diff
@@ -0,0 +1,68 @@
...
+* Membership lines editable
+* Do not calculate membership state from invoice status
+* Start date of last membership period
+
+
+Configuration
+=============
+
+To configure this module, you need to:
```
pedrobaeza: Include here the explanation of where you can configure membership categories.

membership_extension/README.rst
pedrobaeza: Explain for what is used a category.

membership_extension/README.rst
```diff
@@ -0,0 +1,68 @@
...
+
+#. Go to ...
+
+Usage
+=====
+
+To use this module, you need to:
```
pedrobaeza: Include where you can view lines and so on.

membership_extension/models/account_invoice.py
```diff
@@ -0,0 +1,42 @@
...
+            'state': 'waiting',
+        })
+        return super(AccountInvoice, self).action_cancel_draft()
+
+    @api.multi
+    def action_cancel(self):
+        self.mapped('invoice_line').mapped('membership_lines').write({
+            'state': 'canceled',
+        })
+        # TODO: If this is a refund invoice of an invoice related to
```
pedrobaeza: Don't leave TODOs

membership_extension/models/account_invoice.py
```diff
@@ -0,0 +1,42 @@
...
+        # memberbership lines, restore them
+        return super(AccountInvoice, self).action_cancel()
+
+    @api.multi
+    def invoice_validate(self):
+        self.mapped('invoice_line').mapped('membership_lines').write({
+            'state': 'invoiced',
+        })
+        # TODO: If this is a refund invoice of an invoice related to
```
pedrobaeza: Don't leave TODOs

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,79 @@
...
+class MembershipLine(models.Model):
+    _inherit = "membership.membership_line"
+    _order = 'date_to desc, id desc'
+
+    category_id = fields.Many2one(
+        string="Category", comodel_name='membership.membership_category',
+        related="membership_id.membership_category_id", readonly=True)
+    date_from = fields.Date(readonly=False)
+    date_to = fields.Date(readonly=False)
+    state = fields.Selection(
```
pedrobaeza: Do you need to fully overwrite this field?

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,79 @@
...
+        - Non Member: A member who has not applied for any membership.
+        - Cancelled Member: A member who has cancelled his membership.
+        - Old Member: A member whose membership date has expired.
+        - Waiting Member: A member who has applied for the membership and
+                          whose invoice is going to be created.
+        - Invoiced Member: A member whose invoice has been created.
+        - Paid Member: A member who has paid the membership amount."""
+    )
+
+    def __init__(self, pool, cr):
```
pedrobaeza, line 31: Do it in `_register_hook` method and using `patch_method` Odoo's method.

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,79 @@
...
+                        line.date_to = date_to
+
+    @api.onchange('membership_id')
+    def _onchange_membership_id(self):
+        self.member_price = self.membership_id.lst_price
+        self.theorical_dates_set()
+
+    @api.onchange('date')
+    def _onchange_date(self):
+        self.theorical_dates_set()
```
pedrobaeza: Why not include here the code of `theorical_dates_set` and call this method from the other change? I see it more readable.

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @api.multi
+    def _compute_state(self):
```
pedrobaeza, line 57: Why these empty methods? Anyway, it's better to put `pass`

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @api.multi
+    def _inverse_state(self):
+        return True
+
+    @api.multi
+    def unlink(self):
+        model = self.env.context.get('params', {}).get('model', False)
+        if model == 'res.partner':
+            for line in self:
```
pedrobaeza: Save lines: `for line in self.filtered('account_invoice_id'):`

membership_extension/models/res_partner.py
```diff
@@ -0,0 +1,200 @@
...
+        compute="_compute_membership_date",
+        help="Date on which membership has been cancelled.")
+    membership_category_ids = fields.Many2many(
+        string="Membership categories", readonly=True, store=True,
+        comodel_name='membership.membership_category',
+        compute="_compute_membership_state")
+    membership_categories = fields.Char(
+        string="Membership categories", readonly=True, store=True,
+        compute='_compute_membership_state')
+    membership_state = fields.Selection(
```
pedrobaeza, line 47: Do you need to fully overwrite this?

membership_extension/models/res_partner.py
```diff
@@ -0,0 +1,200 @@
...
+        - Old Member: A member whose membership date has expired.
+        - Waiting Member: A member who has applied for the membership and
+                          whose invoice is going to be created.
+        - Invoiced Member: A member whose invoice has been created.
+        - Paid Member: A member who has paid the membership amount."""
+    )
+
+    def __init__(self, pool, cr):
+        super(ResPartner, self).__init__(pool, cr)
+        fields = {'membership_start', 'membership_stop', 'membership_cancel',
```
pedrobaeza, line 54: Same for `_register_hook` and so on.

membership_extension/tests/test_something.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp.tests.common import TransactionCase
+
+
+class SomethingCase(TransactionCase):
```
pedrobaeza: Why this file?

pedrobaeza (changes requested):

membership_autoextend/tests/test_membership_autoextend.py
antespi: This is no longer neccesary

membership_extension/models/membership_line.py
antespi: ```_register_hook``` doesn't work

membership_autoextend/tests/test_membership_autoextend.py
pedrobaeza: OK

membership_extension/models/membership_line.py
antespi: In order to make this field not computed, but regular

membership_extension/models/membership_line.py
pedrobaeza: Why not?

membership_extension/models/membership_line.py
pedrobaeza: Please put a comment about this nasty trick then (you are just omitting the write here in the method as I see, isn't it?). Isn't the field reset when entering the method?

membership_extension/models/membership_line.py
antespi: because modules are not loaded (and then _register_hook called) every time a new addon is installed when you install a list of addons, for instance, in travis test.
Error appears when sale XML data for demo is loaded

pedrobaeza: I have already answered to that: you have to call it manually in your test, like in https://github.com/OCA/server-tools/blob/8.0/base_suspend_security/tests/test_base_suspend_security.py#L28.

It's the only drawback. The rest is better handled this way.

membership_extension/models/res_partner.py
antespi: Done

rafaelbn:
Thanks @antespi , first functional test looks fine 👍  I would like to review again once Pedro's comments are attended.

rafaelbn:
@antespi after last changes, travis fails again

antespi:
cc @Tecnativa

membership_extension/README.rst
```diff
@@ -0,0 +1,72 @@
...
+Then go to membership products and set a category to each one
+
+Usage
+=====
+
+Membership lines are created when a membership product is invoiced, like in
+Odoo standard version. But now users can create a new membership line without
+creating an invoice.
+
+Also, users can edit membership line dates and state even an invoice is related
```
pedrobaeza: `even if an invoice is not related with it`

membership_extension/README.rst
```diff
@@ -0,0 +1,72 @@
...
+====================
+Membership extension
+====================
+
+This module extends Odoo's membership management with the following features:
+
+* Membership Category
+* Membership lines editable
+* Do not calculate membership state from invoice status
+* Start date of last membership period
```
pedrobaeza, line 14: Explain what a category means: `Membership categories allow to classify memberships by types, allowing a partner to be member or not of the different categories.`

membership_extension/README.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Usage
+=====
+
+Membership lines are created when a membership product is invoiced, like in
+Odoo standard version. But now users can create a new membership line without
+creating an invoice.
+
+Also, users can edit membership line dates and state even an invoice is related
+with it.
```
pedrobaeza: Include the text:
```
You will see a general membership status at partner level that specifies if it's a member of any category or not, and also a detail status per membership category.
```

membership_extension/README.rst
pedrobaeza: Include also some information about refund management.

membership_extension/models/membership_category.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields
+
+
+class MembershipCategory(models.Model):
+    _name = "membership.membership_category"
+    _description = "Membership category"
+
+    name = fields.Char(required=True, translate=True)
```
pedrobaeza, line 12: Maybe you can add here a one2many of the membership lines as support for other operations or as query.

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from datetime import timedelta
+from openerp import models, fields, api, _
+from openerp.exceptions import Warning as UserError
+from openerp.addons.membership.membership import STATE
```
pedrobaeza: You should protect this import.

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,73 @@
...
+        string="Category", comodel_name='membership.membership_category',
+        related="membership_id.membership_category_id", readonly=True)
+    date_from = fields.Date(readonly=False)
+    date_to = fields.Date(readonly=False)
+    state = fields.Selection(
+        selection=STATE, readonly=False,
+        compute='_compute_state', inverse="_inverse_state",
+    )
+
+    def __init__(self, pool, cr):
```
pedrobaeza, line 31: Why aren't you still using `_register_hook` method?

membership_extension/models/membership_line.py
```diff
@@ -0,0 +1,73 @@
...
+        model = self.env.context.get('params', {}).get('model', False)
+        action_id = self.env.context.get('params', {}).get('action', False)
+        active_model = self.env.context.get('active_model', False)
+        if not model and action_id:
+            action = self.env['ir.actions.act_window'].browse(action_id)
+            model = action.res_model
+        if not model and active_model:
+            model = active_model
+        if model == 'res.partner' and self.filtered('account_invoice_id'):
+            raise UserError(
```
pedrobaeza, line 68: Can't you put an ondelete='restrict' in any place to avoid this ugly hack?

membership_extension/models/res_partner.py
```diff
@@ -0,0 +1,209 @@
...
+        """Inherit this method to define membership states
+
+        List of membership line states that define a partner as member
+        """
+        return ('invoiced', 'paid')
+
+    def _membership_state_prior(self):
+        """Inherit this method to define membership state precedence
+
+        Diccionary with precendence of each state
```
pedrobaeza: s/Diccionary/Dictionary

pedrobaeza (changes requested):

membership_extension/models/membership_line.py
antespi: Because it doesn't work

membership_extension/models/membership_line.py
antespi: Then member_lines couldn't be removed when invoice line is removed, isn't it?

membership_extension/models/membership_line.py
pedrobaeza: But this is not multi-DB aware. It should work with the other formula.

membership_extension/models/membership_line.py
pedrobaeza: I don't get you. My goal is to avoid this check here.

membership_extension/models/membership_category.py
antespi: This is the same use case of ```res.partner.category``` where we have a many2many field related with partner.
We can define two, one to partners and one to product.template.
But I think we should think about the views and queries before create two more tables in DB, and then this could be an improvement in the future

membership_extension/models/membership_line.py
antespi: Done

membership_extension/models/membership_category.py
pedrobaeza: one2many fields are not tables in DB, just links.

antespi: But this relation is may2many

pedrobaeza: OK, if there's no possibility of one2many, let it this way.

membership_extension/models/membership_line.py
antespi: I tried with _register_hook, but it doesn't work because it is not called always between addons installations.
And I don't known why do you say that this is not multi-DB aware, because there is a pool per environment and one environment per database, so this __init__ is called always when this addon is loaded per each environtment.

membership_extension/models/membership_line.py
antespi: There are two use cases:
1. User removes an account.invoice.line, then membership_lines related has to be removed. There is a method to do it in standard membership addon.
2. User removes a membership_line from res.partner form, This can be done only if this membership_line hasn't an invoice line set.

If I define ```ondelete='restrict'``` and raise and exception in unlink method then I will cover those two use cases?

membership_extension/models/membership_line.py
pedrobaeza: OK, this is not the exact case I faced as I monkey-patched a Python function, but you're working on a pool.

membership_extension/models/membership_line.py
pedrobaeza: OK, we can't use ondelete because there isn't a many2one where we can restrict, but you should use another technique to make this easy and avoid ugly hacks: launch the exception in this method when there's any invoice line and there's no `allow_membership_line_unlink=True` key.

Then, you only need to overwrite the method in `membership` module to call super with this key and you will have both cases correctly covered (and even more, you won't get unexpected membership lines unlinks from other places).

rafaelbn (approved):
Tested in runbot

membership_extension/models/membership_line.py
antespi: Done

pedrobaeza (approved):

PR #23, opened 2017-02-09
[FIX] membership_extension: Refunds cancel

pedrobaeza:
* Avoid ensure_one errors.
* Don't do anything when there's no origin (refund made from scratch or field erased).
* Use more accurate origin field (if refund link module installed)

cc @Tecnativa

### DIFF
--- a/membership_extension/models/account_invoice.py
+++ b/membership_extension/models/account_invoice.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# Copyright 2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openerp import models, api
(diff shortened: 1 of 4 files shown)


membership_extension/models/account_invoice.py
```diff
@@ -18,23 +19,42 @@ def action_cancel_draft(self):
...
-                lines.filtered(lambda r: r.state == 'canceled').write({
-                    'state': origin_state,
-                })
-                lines.write({
-                    'date_cancel': False,
-                })
+        refunds = self.filtered(lambda r: (
+            r.type == 'out_refund' and
+            # TODO: Rename this to `origin_invoice_ids` on v9
+            (r.origin or getattr(r, 'origin_invoices_ids', False))
```
yajo, line 34: As a rule of thumb, you should avoid using `getattr` to access fields, [as you can see in slide 41 of the safe coding tips presentation given in last Odoo Experience](https://www.odoo.com/es_ES/slides/slide/10-rules-for-safer-code-398).

Normally it should be `'origin_invoices_ids' in r and r.origin_invoices_ids`.

However, it does not seem a blocker, since the field name is hardcoded anyway 😉

membership_extension/models/account_invoice.py
```diff
@@ -18,23 +19,42 @@ def action_cancel_draft(self):
...
+        ))
+        for refund in refunds:
+            # Search for the original invoice it modifies
+            if hasattr(refund, 'origin_invoices_ids'):
+                origins = refund.origin_invoices_ids
+            else:
```
yajo, line 40: A try-except would be cleaner IMHO, again a matter of taste.

yajo (approved):
Since it's not much obvious what are you fixing here, I think a test would be good here too. Not a blocker IMO.

membership_extension/models/account_invoice.py
pedrobaeza: The security rule is for injecting names, but not for hardcoded ones, so it doesn't apply here. Also `'origin_invoices_ids' in r` works in this version, but there's no guarantee that future ORM will expose this interface.

pedrobaeza:
@Yajo, I'm not changing anything in the logic, just avoiding errors in it, so tests are already there. Please read the first comment explaining the changes.

pedrobaeza:
@rafaelbn @cubells tests fixed (although not related with the current changes).

pedrobaeza:
Codecov is not correct right now

luismontalba:
@pedrobaeza , please explain what to test here.

cubells:
:+1: 
LGTM

Review code and tested.

pedrobaeza:
@luismontalba no problem, there's already enough reviews. Anyway, the first comment explains what is modified, but you need to know about membership flow. Next time...

PR #32, opened 2017-06-20
[MIG] membership_variable_period: Migration to 10.0

chienandalu:
Depends on:

- [x] membership_extension

cc @Tecnativa


luismontalba:
Please, explain usage or how to review

chienandalu:
@luismontalba From the readme: :blush: 
- Define a member product
- Select 'Variable periods' in the field Membership type. 
- You will be able to select the period for the membership in quantity and interval.

pedrobaeza (approved):
@cubells please review

membership_variable_period/models/account_invoice.py
```diff
@@ -1,37 +1,27 @@
...
-            order="date_to desc")
-        if memb_lines and memb_lines[0].date_to:
-            date_from = (fields.Date.from_string(memb_lines[0].date_to) +
-                         timedelta(days=1))
-        else:
-            date_from = fields.Date.from_string(
-                invoice.date_invoice or fields.Date.today())
-        date_to = (product.product_tmpl_id._get_next_date(date_from) -
+    def _prepare_membership_line(self, invoice, product, price_unit, line_id,
+                                 qty=1):
```
cubells: It must be `qty=1.0` if later you call the method passing a float.

membership_variable_period/views/product_template_views.xml
```diff
@@ -1,74 +1,45 @@
...
+<record model="ir.ui.view" id="membership_products_form_period">
+    <field name="name">Membership Products (variable period)</field>
+    <field name="model">product.template</field>
+    <field name="inherit_id" ref="membership.membership_products_form" />
+    <field name="arch" type="xml">
+        <label for="membership_date_from" position="before">
+            <field name="membership_type"/>
+            <label for="membership_interval_qty"
+                   string="Interval"
+                   attrs="{'invisible': [('membership_type', '!=', 'variable')]}"/>
```
cubells, line 16: It would be interesting to add parameter `t-translation="off"` in order to avoid attributes be translated, isn't it?

membership_variable_period/views/product_template_views.xml
```diff
@@ -1,74 +1,45 @@
...
+            <div attrs="{'invisible': [('membership_type', '!=', 'variable')]}">
+                <field name="membership_interval_qty"
+                       attrs="{'required': [('membership_type', '=', 'variable'), ('membership', '=', True)]}"
```
cubells, line 19: Same here

membership_variable_period/views/product_template_views.xml
```diff
@@ -1,74 +1,45 @@
...
+                       class="oe_inline"
+                       nolabel="1"/>
+                <field name="membership_interval_unit"
+                       attrs="{'required': [('membership_type', '=', 'variable'), ('membership', '=', True)]}"
```
cubells, line 23: Same here

cubells (changes requested):
Few changes to approve

membership_variable_period/views/product_template_views.xml
chienandalu: Aren't the second values of the selection tuples the ones being translated (like so: https://github.com/Tecnativa/vertical-association/blob/19ea87acbf6bb4adae7c1f522fa97ddc192d4065/membership_variable_period/models/product_template.py#L39)

Anyway, isn't `t-translation` only valid for qweb templates?

chienandalu:
Seems that the last rebase I made messed up history :confused:

pedrobaeza:
@cubells please answer

membership_variable_period/views/product_template_views.xml
cubells: It's possible @chienandalu . I have not tested it.

But anyway I recreated the translations files for that module and these strings: 
```
msgid "{'invisible': [('membership_type', '!=', 'fixed')]}"
```
and these
```
msgid ""
"{'required': [('membership_type', '=', 'fixed'), ('membership', '=', True)],"
" 'invisible': [('membership_type', '!=', 'fixed')]}"
msgid "{'invisible': [('membership_type', '!=', 'fixed')]}"
```
and these:
```
msgid ""
+"{'required': [('membership_type', '=', 'fixed'), ('membership', '=', True)],"
+" 'invisible': [('membership_type', '!=', 'fixed')]}"
```
are not created.
So you have to avoid to be translated or you have an incorrect po files.

chienandalu: @cubells I see now. Fixed in the .po files. Weird thing anyway. I guess it's been there from long time ago. I tried to export the module's translations and got it right.

cubells (approved):
:+1: 

LGTM

Code review and test

pedrobaeza:
Please squash for merging

chienandalu:
@pedrobaeza Done

PR #38, opened 2017-09-28
[ADD] membership_delegated_partner: New module 10.0

chienandalu:
Membership Delegated Partner
============================

This module allows to delegate de membership line to a partner independently of
the invoicing partner.

cc @Tecnativa

### DIFF
--- a/membership_delegated_partner/README.rst
+++ b/membership_delegated_partner/README.rst
@@ -1,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Membership Delegated Partner
+============================
+
+This module allows to delegate a membership line to a partner independently of
+the invoicing partner.
+
+Configuration
+=============
+
+In *Settings > Users > <your_user>* set *Delegated partner in membership*
+permission on.
+
+Usage
+=====
+
+#. In an invoice with membership lines, choose a delegated partner.
+#. The membership line will go to the delegated partner.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/vertical-association/issues>`_. In case of trouble,
+please check there if your issue has already been reported. If you spotted it
+first, help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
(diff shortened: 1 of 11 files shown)


chienandalu:
@rafaelbn Please review

rafaelbn (approved):
Super, tested in runbot 👍

chienandalu:
~~@rafaelbn @pedrobaeza Do you think a security group is convenient here?~~

Ok, I just added it. There are going to be cases in which only a set of users should be allowed to force this change.

pedrobaeza:
I don't see that you need a security group for this. If you don't want this feature, don't install the module, but I can't think of any case where some people is allowed to put a delegate and other don't. Please check also Travis status.

membership_delegated_partner/models/account_invoice.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    delegated_member = fields.Many2one(
```
pedrobaeza: delegated_member_id (it's a many2one)

chienandalu:
@pedrobaeza Well, this module gives the power to change the associated member to any user with accounting permissions. It wouldn't be rare to wish to restrict that to just a few users.

pedrobaeza:
Check Travis status. OK about the group.

chienandalu:
No runbot here either :confused:

pedrobaeza (approved):

membership_delegated_partner/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Membership Delegated Partner
+============================
+
+This module allows to delegate de membership line to a partner independently of
```
pedrobaeza: s/de/a

chienandalu:
@pedrobaeza fixed

pedrobaeza:
@luismontalba @rafaelbn please review

rafaelbn:
I agree with https://github.com/OCA/vertical-association/pull/38#issuecomment-333315164

I don't get this point: https://github.com/OCA/vertical-association/pull/38#issuecomment-333466751

But not blocking.

Tested again 👍

pedrobaeza:
@chienandalu Migrate it to v11

PR #39, opened 2017-09-29
[ADD] contract_membership_delegated_partner: New module 10.0

chienandalu:
Contract Membership Delegated Partner
=====================================

This module allows to set the delegated membership at the contract so the
recurring invoices automatically have it set.

Depends on:

- [x] membership_delegated_partner https://github.com/OCA/vertical-association/pull/38

cc @Tecnativa

### DIFF
--- a/contract_membership_delegated_partner/models/account_analytic_account.py
+++ b/contract_membership_delegated_partner/models/account_analytic_account.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = 'account.analytic.account'
+
+    delegated_member_id = fields.Many2one(
+        comodel_name='res.partner',
+        string='Delegated Member',
+    )
+
+    def _prepare_invoice(self):
+        res = super(AccountAnalyticAccount, self)._prepare_invoice()
+        res['delegated_member_id'] = self.delegated_member_id.id
+        return res
--- a/contract_membership_delegated_partner/tests/test_delegated_partner_contract.py
+++ b/contract_membership_delegated_partner/tests/test_delegated_partner_contract.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.contract.tests.test_contract import TestContractBase
+
+
+class TestMembershipDelegateSetup(TestContractBase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestMembershipDelegateSetup, cls).setUpClass()
+        cls.partner2 = cls.env['res.partner'].create({
+            'name': 'Mrs. Odoo',
+        })
+        cls.product.membership = True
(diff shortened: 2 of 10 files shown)


rafaelbn:
WIP until https://github.com/OCA/vertical-association/pull/38 is merged thanks!

contract_membership_delegated_partner/models/account_analytic_account.py
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = 'account.analytic.account'
+
+    delegated_member = fields.Many2one(
```
pedrobaeza: delegated_member_id.

chienandalu:
@pedrobaeza Changes done

pedrobaeza:
Check Travis status

chienandalu:
@pedrobaeza Depends on PR https://github.com/OCA/vertical-association/pull/38 wich is not merged yet

chienandalu:
@pedrobaeza @rafaelbn Ready to review (if runbot lets it)

pedrobaeza:
Please rebase to have the dependent module

chienandalu:
@pedrobaeza @rafaelbn All green now

contract_membership_delegated_partner/tests/test_delegated_partner_contract.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestMembershipDelegateContract(common.SavepointCase):
```
pedrobaeza: Can't you inherit from base contract test? What I usually do is to split a base test class in 2:

* One base class that inherits from `common.SavepointCase` which implements the `setUpClass` method.
* A second class that inherits from the previous base one, which implements the tests.

Then, on this module, inherit from base class and add only the records you need for the specific case. Well, for this specific case the setup is not very large, but imagine the same with other situations.

chienandalu: @pedrobaeza Do you have an example? I can't figure out how to reach the contract tests route so I can inherit the class.

pedrobaeza: `from odoo.addons.contract.tests.<test_file> import <test_class_name> `

chienandalu:
@pedrobaeza Test changed

pedrobaeza:
You haven't understood me well:

- First you need to modify contract tests for having the dual classes. I have done it in this PR: https://github.com/OCA/contract/pull/105
- Then you need to adapt tests of this module for minimizing the new created objects and only checks the things that correspond to the module. See 7fece17444b67012e3165ec96ab076284b938c01 with the reduce test.

Unfortunately, tests are failing on contract side due to another bug in main branch. We will restart this when the problem is fixed.

pedrobaeza:
Relaunching now that OCA/contract#105 has been merged.

rafaelbn (approved):
Tested functionaly in runbot 👍 

But travis is red, please @chienandalu  review travis. Thanks!

pedrobaeza:
@rafaelbn Travis was red due to some things I forgot to change in the redone tests. Now it's green and I'll merge after my review and yours.

PR #41, opened 2017-10-03
[MIG] membership_prorate: Migration to 11.0

chienandalu:
Prorate the membership fee
===========================

cc @Tecnativa


membership_prorrate/README.rst
```diff
@@ -0,0 +1,62 @@
+===========================
+Prorrate the membership fee
```
lepistone: Are you sure about the word `prorrate`? What I have heard is `pro rata`. I'm not a native English speaker. Can you investigate please?

membership_prorrate/models/account_invoice.py
```diff
@@ -0,0 +1,61 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import models, api, fields
+from datetime import timedelta
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = "account.invoice.line"
+
+    def _get_membership_interval(self, product, date):
```
lepistone: How is `date` used here?

membership_prorrate/models/account_invoice.py
```diff
@@ -0,0 +1,61 @@
...
+        return date_from, date_to
+
+    def _prepare_invoice_line_prorrate_vals(self, invoice_line):
+        product = invoice_line.product_id
+        date_invoice = fields.Date.from_string(
+            invoice_line.invoice_id.date_invoice or fields.Date.today())
+        date_from, date_to = self._get_membership_interval(
+            product, date_invoice)
+        date_invoice = date_from if date_invoice < date_from else date_invoice
+        date_invoice = date_to if date_invoice > date_to else date_invoice
```
lepistone: Suggestion to make it more explicit:

```python
if date_invoice < date_from:
    date_invoice = date_from
if date_invoice > date_to:
    date_invoice = date_to
```

membership_prorrate/models/account_invoice.py
```diff
@@ -0,0 +1,61 @@
...
+        product = self.env['product.product'].browse(
+            vals.get('product_id', False)
+        )
+        if not product.membership or not product.membership_prorrate:
+            return invoice_line
+        # Change quantity accordingly the prorrate
+        invoice_line_vals = self._prepare_invoice_line_prorrate_vals(
+            invoice_line)
+        if invoice_line_vals:
+            date_from = invoice_line_vals.pop('date_from')
```
lepistone: This removes `date_from`from the dict. Is that expected?

membership_prorrate/tests/test_membership_prorrate.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_create_invoice_membership_product_wo_prorrate(self):
+        self.product.membership_prorrate = False
+        invoice_id = self.partner.create_membership_invoice(
+            product_id=self.product.id, datas={'amount': 1.0})
+        invoice = self.env['account.invoice'].browse(invoice_id)
+        self.assertEqual(invoice.invoice_line_ids[0].quantity, 1.0)
+
+    def test_create_invoice_membership_product_prorrate(self):
+        account = self.partner.property_account_receivable_id.id
```
lepistone: This is an ID. Instead, we could set `account = self.partner.property_account_receivable_id` and use `account.id` later on.

lepistone (approved):
I left some comments, but they are not blocking as far as I'm concerned. Thanks!

pedrobaeza:
@chienandalu can you attend comments here?

membership_prorrate/models/account_invoice.py
chienandalu: Is a parameter ready for inheritance (i.e. membership_prorrate_variable_period uses it). As the comments state:
```
:param date: date object for the requested date (if needed for
        inherited computations)
```

membership_prorrate/README.rst
chienandalu: Seems that the correct spelling is **prorate**: https://www.merriam-webster.com/dictionary/prorate

pedrobaeza: Then we should rename the module

membership_prorrate/models/account_invoice.py
chienandalu: Yes, as you can see the dict is written into the invoice line in the next code line and that value isn't needed since it's not in the model.

membership_prorrate/models/account_invoice.py
chienandalu: Mmm, isn't this just a matter of preference :slightly_smiling_face:

pedrobaeza: It's more readable

chienandalu: @pedrobaeza Wich of the two?

pedrobaeza: The proposed one

pedrobaeza:
Rebase now to see new status (+ coverage which is missing right now).

chienandalu:
All green

PR #42, opened 2017-10-03
[MIG] membership_extension: Migration to 11.0

chienandalu:
Membership extension
====================

This module extends Odoo's membership management with the following features:

* Membership category
* Membership lines editable
* Do not calculate membership state from invoice status
* Start date of last membership period

cc @Tecnativa

PR #44, opened 2017-10-03
[MIG] membership_variable_period: Migration to 11.0

chienandalu:
Variable period for memberships
===============================

Current membership module allows to set products that define a fixed period
for a membership. This is good when the quotas are defined periodically, and
when you become a member, you are until the end of that quota cycle.

But a lot of times, membership quotas express an amount of time that you
gain the membership. For example, one year since the subscription.

This module allows to make it in Odoo, using current membership features,
and adapting them for this purpose. As now the quota is not attached to a fixed
period, you can also invoice more than one quantity for being a member for
the corresponding number of periods.

Depends on:
- [x] membership_extension: https://github.com/OCA/vertical-association/pull/42 

cc @Tecnativa


pedrobaeza:
Can we get full coverage?

rafaelbn (approved):
Functionally tested :+1:

chienandalu:
All green, full coverage and rebased

pedrobaeza (approved):

PR #51, opened 2017-12-13
[10.0][IMP] membership_extension: adhered member

chienandalu:
- When a partner has an associated member there is the option to make member
  start date independent from the associated member one so joining dates can be
  tracked for those members.

- To make member start date independent from the associated member one, check the
  option *Adhered member* in the membership tab of the partner who is associating.
  A start date will now be available to edit.

cc @Tecnativa

### DIFF
--- a/membership_extension/models/res_partner.py
+++ b/membership_extension/models/res_partner.py
@@ -21,6 +21,16 @@ class ResPartner(models.Model):
     _inherit = "res.partner"
 
     associate_member = fields.Many2one(index=True)
+    is_adhered_member = fields.Boolean(
+        string="Adhered member",
+        help="A member who is associated to another one, but whose membership "
+             "are independent.",
+    )
+    membership_start_adhered = fields.Date(
+        string="Membership Start Date",
+        help="Date from which partner is adhered.",
+        default=fields.Date.today(),
+    )
     membership_start = fields.Date(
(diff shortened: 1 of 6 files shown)


rafaelbn (approved):
Tested functionally in customer enviroment @chienandalu 👍 

Please add me as a contributor: losts of hour of myself 😄

membership_extension/models/res_partner.py
```diff
@@ -21,6 +21,16 @@ class ResPartner(models.Model):
     _inherit = "res.partner"
 
     associate_member = fields.Many2one(index=True)
+    is_adhered_member = fields.Boolean(
+        string="Adhered member",
+        help="A member who is associated to another one, but whose membership "
+             "are independent.",
```
cubells, line 27: Maybe with this new feature, you must change help of associate_member field, because now the date is independent. Am I right?

cubells (approved):
👍 

LGTM

review code and testing.

luismontalba (approved):
Tested in local

membership_extension/models/res_partner.py
chienandalu: Not really. What an adhered member sees is the field `membership_start_adhered` wich will written onto it's `membership_start` when computed. And that field has its own help (*Date from which partner is adhered*)

PR #52, opened 2017-12-14
[10.0][FIX] membership_extension: method-super mismatch

chienandalu:
There was a mismatch between a method and its super call that caused unexpected behaviour on other modules using the method.

cc @Tecnativa

### DIFF
--- a/membership_extension/__manifest__.py
+++ b/membership_extension/__manifest__.py
@@ -7,7 +7,7 @@
 {
     "name": "Membership extension",
     "summary": "Improves user experience of membership addon",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
hehe, no comments :stuck_out_tongue_closed_eyes: 

Check the TODO comment that is a couple of lines below to see if it can be removed.

chienandalu:
yeah... :sweat_smile: 

Yes, the TODO seems to be old. Updated docstring also.

pedrobaeza (approved):
Fast-tracking due to the critical bug

PR #53, opened 2017-12-20
[10.0][fix] membership_extension: add security group to menu

luismontalba:
- Add access right to the main menu "Members" to employees group, avoiding restriction for only accountants when the association module is installed.

### DIFF
--- a/membership_extension/__manifest__.py
+++ b/membership_extension/__manifest__.py
@@ -8,7 +8,7 @@
 {
     "name": "Membership extension",
     "summary": "Improves user experience of membership addon",
-    "version": "10.0.1.1.4",
+    "version": "10.0.1.2.0",
     "category": "Membership",
     "author": "Tecnativa, "
               "Odoo Community Association (OCA)",
--- a/membership_extension/__manifest__.py
+++ b/membership_extension/__manifest__.py
@@ -20,9 +20,10 @@
         "membership",
     ],
     "data": [
-        "security/ir.model.access.csv",
         "security/membership_security.xml",
+        "security/ir.model.access.csv",
         "views/membership_category_view.xml",
+        "views/membership_views.xml",
         "views/product_template_view.xml",
         "views/res_partner_view.xml",
         "data/membership_category_data.xml",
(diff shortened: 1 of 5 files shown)


pedrobaeza:
This is not security. It's changing a menu field, so it should go in /views/.

Any way, this is not what we have talked. I want a new security group for granting specifically that. And check all ACLs needed to that group. Imagine you only have that new group, and you need to access the same objects in that menu than any other role with that permission.

chienandalu:
@pedrobaeza I've made the security group, and gave permissions so a user with just this group could access to the whole features of the  membership application.

membership_extension/__manifest__.py
```diff
@@ -2,12 +2,13 @@
...
 # Copyright 2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
+# Copyright 2017 Luis M. Ontalba <luis.martinez@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     "name": "Membership extension",
     "summary": "Improves user experience of membership addon",
-    "version": "10.0.1.1.3",
+    "version": "10.0.1.1.4",
```
pedrobaeza: Change version to 10.0.1.2.0

membership_extension/security/membership_security.xml
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+
+    <record model="ir.module.category" id="module_category_membership_management">
+        <field name="name">Membership</field>
+        <field name="description">Helps you manage operations for managing memberships.</field>
+        <field name="sequence">3</field>
+    </record>
+
+    <record id="group_membership_manager" model="res.groups">
+        <field name="name">Membership manager</field>
```
pedrobaeza: Call it only `Manager`, as the display name will be `Membership / Manager`

pedrobaeza (changes requested):
Are you sure you don't need more permissions?

chienandalu:
@pedrobaeza Changes done (added one ACL more) and tested that a user with just **Membership manager**, **Employee** and **Contact creation** permissions can do all member operations except billing (for wich the **Billing** permission would be granted)

pedrobaeza (approved):

rafaelbn (approved):
Functionally tested in runbot 👍

rafaelbn:
Thanks @chienandalu

PR #55, opened 2018-02-27
[FIX] membership_initial_fee: getattr fails

chienandalu:
`getattr()` fails if the attribute doesn't exist. We use `hasattr()` instead.

cc @Tecnativa


pedrobaeza:
Change module version number

chienandalu:
done

pedrobaeza (approved):
Fast-tracking this

PR #57, opened 2018-03-29
[IMP] membership_extension: Adapt to standard refund field

pedrobaeza:
This module uses in previous versions an indirect dependency to account_invoice_refund_link for detecting the origin invoice where the refund comes from, but now there's a field in standard for
pointing to that origin invoice, so no need of such code.

@Tecnativa

### DIFF
--- a/membership_extension/models/account_invoice.py
+++ b/membership_extension/models/account_invoice.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
-# Copyright 2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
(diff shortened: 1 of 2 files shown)


luismontalba (approved):
Tested on runbot

membership_extension/models/account_invoice.py
```diff
@@ -22,39 +22,24 @@ def action_invoice_draft(self):
...
-                    origin_state = (
-                        'paid' if origin.state == 'paid' else 'invoiced'
-                    )
-                    lines.filtered(lambda r: r.state == 'canceled').write({
-                        'state': origin_state,
-                    })
-                    lines.write({
-                        'date_cancel': False,
-                    })
+            origin = refund.refund_invoice_id
```
chienandalu, line 33: why do assign it to a variable?

pedrobaeza: For shortening the reference that is later used.

chienandalu: Ok, guessed it was something like that :smile:

chienandalu (approved):

PR #63, opened 2018-10-01
[FIX] initial_fee_create_check call site

j-zaballa:
initial_fee_check_create has to be called with the newly created invoice
line, not with self. Self is just the model in the create method from
which initial_fee_check_create is called.

### DIFF
--- a/membership_initial_fee/models/account_invoice.py
+++ b/membership_initial_fee/models/account_invoice.py
@@ -3,7 +3,8 @@
 # Copyright 2017 Tecnativa - David Vidal
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import models, api, _
+from odoo import api, models
+from odoo.tools.translate import _
 
 
 class AccountInvoiceLine(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Well, in this case, `self` is not being used, so you can mark the model as `@api.model` for making this significant. Or do you need that created line for anything? If you make the change, then better remove `product` argument and take `self.product`.

j-zaballa:
I think `self` needs to be used to check if a given partner needs to be charged the initial fee or not. We need to call the method with the newly created invoice line, so self holds relevant information when looking for member lines in `initial_free_create_check`.
And yes, you're right about the product argument. I missed that one :)

membership_initial_fee/models/account_invoice.py
```diff
@@ -32,11 +32,11 @@ def _prepare_initial_fee_vals(self):
         return inv_line._convert_to_write(inv_line._cache)
 
     @api.multi
-    def initial_fee_create_check(self, product):
```
chienandalu, line 35: Mmm, I get your point here, but you should keep compatibility with other possible modules using/extending this method.

Something like:
```python
def initial_fee_create_check(self, product=False):
    # TODO: remove product parameter in v12
    product = product or self
    if not product or not product.membership or product.initial_fee == 'none':
        return False
```

j-zaballa: Yes, It makes total sense. Thanks for the feedback. I'll push the changes according to your suggestion.

pedrobaeza:
I still prefer to use self.product and not to add the argument

chienandalu:
Sure, I just gave @j-zaballa a hint so we don't break possible current version uses while moving to that behaviour already.

pedrobaeza:
I still don't understand why why need the extra argument instead of getting it from the line. And Travis is red.

j-zaballa:
We don't need it, but someone who is extending/using the module and calling the method with the extra product argument might need it. That is what @chienandalu is saying and I for one agree with him. I think Travis is red because of a code line being too long. I'll fix it right away.

pedrobaeza:
No, @chienandalu says that he prefers also my option, but explain the compatible way to do it, but you don't need that for extending the use, as you can always overwrite the method.

Summarizing, don't change this and use `self.product_id`.

j-zaballa:
Of course, I also prefer to remove the argument. IMHO the compatible way to do it is not to remove it yet, because someone might be calling the method with the product argument (e.g. from another module that depends on this one). Removing the argument would break that piece of code.

pedrobaeza:
OK, I see now that the argument was before. Sorry for my confusion. Yes, we should keep it until next version.

pedrobaeza:
Have you signed CLA?

j-zaballa:
I sure did :)

pedrobaeza (approved):

PR #65, opened 2019-05-20
[ADD] add module membership_product_set

yweng8111:
This module enable the user to define a Set of Memberships and create the summarized invoice.


chienandalu (commented):
I wonder if the same could be achieved via product.bom kits

yweng8111:
> I wonder if the same could be achieved via product.bom kits

Thanks for the Review. Did you mean that I should use mrp.bom? I have also considered about it. But I just want to keep it small and independent from other module. Peoples who want to use it, doesn't need to install the module "mrp".

membership_product_set/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
...
+            for partner in self:
+                addr = partner.address_get(['invoice'])
+                if partner.free_member:
+                    raise UserError(_("Partner is a free Member."))
+                w = _("Partner doesn't have an address to make the invoice.")
+                if not addr.get('invoice', False):
+                    raise UserError(w)
+                account_id = partner.property_account_receivable_id.id
+                position_id = partner.property_account_position_id.id
+                invoice = self.env['account.invoice'].create({
```
nikul-serpentcs, line 26: Add Condition `if  account_id and position_id`
Because of Sometime property_account_receivable_id  & property_account_position_id is Deleted then ..

yweng8111: Thanks for the review. I have added the condition "if account_id".
"fiscal_position_id" is not required for account.invoice. So we can just ignore it, if it was not set.

kempleton (commented):
I've tested the functionality and everything works fine.

kempleton (approved):

membership_product_set/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    'name': 'Product Set for memberships',
+    'version': '11.0.1.0.0',
+    'license': 'AGPL-3',
+    'category': 'Association',
+    'author': 'Yu Weng, '
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/vertical-association',
+    'depends': [
+        'membership',
```
yucer, line 13: Where is that module 'membership' defined ? I don't see it in the v11 branch.

yweng8111: It is here.
https://github.com/odoo/odoo/tree/11.0/addons/membership

membership_product_set/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+                        'product_id': product.id,
+                        'price_unit': amount,
+                        'invoice_id': invoice.id,
+                    }
+
+                    # create a record in cache, apply onchange
+                    # then revert back to a dictionnary
+                    invoice_line = self.env['account.invoice.line']\
+                        .new(line_values)
+                    invoice_line._onchange_product_id()
```
yucer, line 44: Why do you need all of this after the explicit call of  **_onchange_product_id** ?

```
                    line_values = invoice_line._convert_to_write({
                        name: invoice_line[name]
                        for name in invoice_line._cache
                    })
                   line_values['price_unit'] = amount
```

It is just needed in this context or always after changing the product_id. In that case it would be better to extend the method **_onchange_product_id** here in this module.

Also if the behavior of changing the name and the price unit is already covered by other onchange, then it might be better not to call explicitly the method of **_onchange_product_id** , note that this method is protected and not guaranteed to work always.

The ideally would be to call the generic *onchange* method like [here](https://github.com/odoo/odoo/blob/11.0/odoo/addons/test_new_api/tests/test_onchange.py#L146) so that Odoo can decide which onchange methods are needed and in which order.

membership_product_set/tests/test_membership.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def test_membership_invoice(self):
+        wizard = self.env['membership.invoice'].with_context(
+            active_ids=[self.partner.id],
+            active_model='res.partner'
+        ).create({
+            'product_id': self.product_set.id,
+            'member_price': 150,
+        })
+        with self.assertRaises(UserError):
```
yucer: Consider using  **assertRaisesRegexp**, it can avoid to catch another **UserError** not intended and swallow the failure. 

` with self.assertRaisesRegexp(UserError, r'^My expected message')`

membership_product_set/models/res_partner.py
yucer: but this is also a problem of the original method that is being extended here.

membership_product_set/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.multi
+    def create_membership_invoice(self, product_id=None, datas=None):
+        product_id = product_id or datas.get('membership_product_id')
+        amount = datas.get('amount', 0.0)
+        product = self.env['product.product'].browse(product_id)
+        invoice_list = []
+        if product.membership_set:
```
yucer, line 17: The original [create_membership_invoice](https://github.com/OCA/OCB/blob/11.0/addons/membership/models/partner.py#L173) has a lot of code in common with this one ?

Try not to copy and paste it just for a certain condition. Try to make a method in the base method that perform the creation of **line_values** and just overwrite that method in this class.

My other suggestions are inherited from the base code being extended, but this is very important for the future maintainability of this method.

yucer (changes requested):
Very good.

Just don't repeat the code in **create_membership_invoice** and extend just the calculation of line_values.

membership_product_set/models/res_partner.py
yweng8111: My extension will only be executed, if product.membership_set is True. Otherwise the origin function will be called. This function can also be inherited by other module. I don't see any problem now. I don't want to change the origin function from Odoo Repo and create a new PR of it.

membership_product_set/models/res_partner.py
yweng8111: Because invoice_line is in that moment only in cache. It is good for the performance.

membership_product_set/tests/test_membership.py
yweng8111: Thanks for the tip. I will improve it.

membership_product_set/models/res_partner.py
yucer: ok

yucer (approved):

rafaelbn (approved):

PR #68, opened 2019-05-24
[MIG] membership_withdrawal: Migration to 12.0

chienandalu:
This module allows to record membership withdrawal reason and date of request.

cc @Tecnativa


chienandalu:
Weird error from runbot:
```
Unable to find image 'oca-vertical-association:0dc8f83_3' locally
docker: Error response from daemon: repository oca-vertical-association not found: does not exist or no pull access.
See 'docker run --help'.
```

pedrobaeza:
This is runbot flickering again..

chienandalu:
It passed at the second try :)

membership_withdrawal/__manifest__.py
```diff
@@ -1,12 +1,11 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
 # Copyright 2017-18 David Vidal <david.vidal@tecnativa.com>
```
pedrobaeza: You can update the year

pedrobaeza (approved):

rafaelbn:
@chienandalu after a credit note and cancelling the association the membership status doesn't change and the res.partner is still member/associated

chienandalu:
@rafaelbn I'd say that problem doesn't fit in this module scope and is more a missing feature in the core membership module.

What I wonder as well is, when a credit note is issued: does it always mean the membership is canceled?

For example:

- A customer is charged with a membership quota and an initial fee.
- Later, our commercial department says that this is a VIP customer and asks accounting to make a refund of the quota keeping the initial fee.

rafaelbn:
Please @chienandalu ADD this important USE CASE in known issue to warn as this problem could be dramatic in a high volume associations

rafaelbn (approved):
Functional review :+1: 

As commented before merging attend https://github.com/OCA/vertical-association/pull/68#issuecomment-499004118

pedrobaeza:
@chienandalu can you add the use case?

chienandalu:
But it's not an issue with this module wich doesn't change the status of the membership at all.

I think that flow fix is more adequate for `membership_extension`

pedrobaeza:
OK, merging this one and please check if we should add it in `membership_extension`.

PR #70, opened 2019-05-24
[MIG] membership_prorate: Migration to 12.0

chienandalu:

cc @Tecnativa


membership_prorate/models/account_invoice.py
```diff
@@ -0,0 +1,61 @@
...
+        date_to = fields.Date.from_string(product.membership_date_to)
+        return date_from, date_to
+
+    def _prepare_invoice_line_prorate_vals(self, invoice_line):
+        product = invoice_line.product_id
+        date_invoice = fields.Date.from_string(
+            invoice_line.invoice_id.date_invoice or fields.Date.today())
+        date_from, date_to = self._get_membership_interval(
+            product, date_invoice)
+        if date_invoice < date_from:
```
nikul-serpentcs, line 29: @chienandalu  here use `If..else...`
```
if date_invoice < date_from:
    date_invoice = date_from
else:
    date_invoice = date_from
```

nikul-serpentcs (commented):
Minor Change

membership_prorate/models/account_invoice.py
chienandalu: Thanks, @nikul-serpentcs Done

chienandalu: Sorry @nikul-serpentcs  , but the change wasn't correct and I revert it.

The purpose of this is to adjust the invoice date inside the membership product period. So the structure is correct:

- The invoice date (e.g.: today) could be **before** that period, thus we would force it to the beginning of that period (the **full qouta** would be applied)
- The invoice date (e.g.: today) could be **after** that period, thus we would force it to the ending of that period (**no qouta** would be applied)
- Or the invoice date could be **inside** that period, , thus we wouldn't change it and the **prorated quota** would be applied.

nikul-serpentcs: @chienandalu Only Code Review, Not tested in Runbot.
If it's not correct then revert it.
Thanks

pedrobaeza (approved):

rafaelbn (approved):
Tested in runbot :+1:

Issue #80, opened 2019-11-21
Membership Category is not multicompany compliant


astirpe:
The module `membership_extension` adds a new model for "Membership Category" but this model is not multicompany aware. So what happens is that, in a multicompany environment, all the categories are accessible by all the companies.

astirpe:
Fixed in V12 and V13

PR #81, opened 2019-11-21
[12.0][IMP] membership_extension: Membership Category multicompany compliant

astirpe:
Implements #80

### DIFF
--- a/membership_extension/models/membership_category.py
+++ b/membership_extension/models/membership_category.py
@@ -1,7 +1,9 @@
 # Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# Copyright 2019 Onestein - Andrea Stirpe
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models, fields
+from odoo import models, fields, api, _
+from odoo.exceptions import ValidationError
 
 
 class MembershipCategory(models.Model):
@@ -9,3 +11,22 @@ class MembershipCategory(models.Model):
     _description = "Membership category"
 
     name = fields.Char(required=True, translate=True)
+    company_id = fields.Many2one('res.company')
+
+    @api.constrains("company_id")
+    def _check_company_id(self):
+        if self.env.context.get("bypass_company_validation"):
+            return
+        categories = self.filtered(lambda c: c.company_id)
+        templates = self.env['product.template'].search([
+            ('membership_category_id', 'in', categories.ids),
+        ]).filtered(
+            lambda t: t.company_id
+            and t.company_id != t.membership_category_id.company_id
+        )
+        if templates:
+            raise ValidationError(_(
+                "You cannot change the Company, as this "
+                "Membership Category is used by Product Template (%s), "
+                "which has an incompatible assigned Company."
+            ) % fields.first(templates).name)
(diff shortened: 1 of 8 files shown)


chienandalu (changes requested):
Looks good, although tests fail. It'd be nice to have this new case tested as well.

astirpe:
@chienandalu tests added. PR green!

astirpe:
@chienandalu would you update your review?

membership_extension/views/product_template_view.xml
```diff
@@ -19,8 +19,8 @@
     <field name="model">product.template</field>
     <field name="inherit_id" ref="membership.membership_products_form" />
     <field name="arch" type="xml">
-        <field name="membership" position="after">
-            <field name="membership_category_id" />
+        <field name="company_id" position="after">
```
chienandalu, line 22: Why changing the position?

astirpe: Because of this onchange: https://github.com/OCA/vertical-association/pull/81/files#diff-7e6c873c7ba99b0bf6d9e8391c33a96fR26-R28

If we keep `membership_category_id` before `company_id`, it results in a poor UX: imagine when the user sets a `membership_category_id, then sets the company and the onchange overwrites the previously selected `membership_category_id`.
To avoid such confusing behavior, I moved the field in a more logical position.

chienandalu: Ok, but the two fields are together and `membership` is invisible anyway :sweat_smile:

chienandalu (approved):

membership_extension/models/product_template.py
```diff
@@ -10,7 +11,9 @@ class ProductTemplate(models.Model):
 
     membership_category_id = fields.Many2one(
         string="Membership category",
-        comodel_name='membership.membership_category')
+        comodel_name='membership.membership_category',
+        domain="['|',('company_id','=',False),('company_id','=',company_id)]",
```
pedrobaeza: This is not needed. Record rule already forces this.

membership_extension/models/membership_category.py
```diff
@@ -1,11 +1,29 @@
...
+
+    @api.constrains("company_id")
+    def _check_company_id(self):
+        if self.env.context.get("bypass_company_validation"):
+            return
+        categories = self.filtered(lambda c: c.company_id)
+        templates = self.env['product.template'].search([
+            ('membership_category_id', 'in', categories.ids),
+        ]).filtered(
+            lambda t: t.company_id != t.membership_category_id.company_id)
```
pedrobaeza: If the company of the template is False, it's correct to let that category to be put.

membership_extension/models/product_template.py
```diff
@@ -19,3 +22,11 @@ def _get_next_date(self, date, qty=1):
             date_to = fields.Date.from_string(self.membership_date_to)
             return date_to + timedelta(1)
         return False  # pragma: no cover
+
+    @api.onchange('company_id')
+    def _onchange_company_id_compute_membership_category_id(self):
+        """Reset the Membership Category in case a different Company is set"""
+        category_company = self.membership_category_id.company_id
+        if self.company_id and category_company:
+            if not category_company == self.company_id:
```
pedrobaeza: Weird way to put it...
```suggestion
            if category_company != self.company_id:
```

membership_extension/models/membership_category.py
```diff
@@ -1,11 +1,29 @@
...
+        if templates:
+            raise ValidationError(_(
+                "You cannot change the Company, as this "
+                "Membership Category is used by Product Template (%s)."
```
pedrobaeza: I think you should expand explanation saying `..., which has an incompatible assigned company.`

pedrobaeza (changes requested):

astirpe:
@pedrobaeza done! Could you update your review?

pedrobaeza (approved):

PR #82, opened 2019-11-22
[13.0][MIG] membership_delegated_partner

astirpe:
Porting to V13.
This PR is also a proposal to use a different approach for calculating the membership lines.
The proposal is to use a compute method for the partner in the membership lines.


chienandalu (approved):
Thanks :+1: :)

pedrobaeza (approved):

PR #90, opened 2020-04-28
[10.0][FIX] membership_variable_period - fixing ensure_one() error on product.template write

### LINKED ISSUE #89: [10.0] membership_variable_period
This:
https://github.com/OCA/vertical-association/blob/81e51d8cae307e94a893d217b7abdf9c79209276/membership_variable_period/models/product_template.py#L67
is bad.
If self contains more than one record it raises a ValueError.

baimont:
fixes https://github.com/OCA/vertical-association/issues/89

### DIFF
--- a/membership_variable_period/models/product_template.py
+++ b/membership_variable_period/models/product_template.py
@@ -63,6 +63,11 @@ class ProductTemplate(models.Model):
 
     @api.multi
     def write(self, vals):
-        self._correct_vals_membership_type(
-            vals, vals.get('membership_type', self.membership_type))
-        return super(ProductTemplate, self).write(vals)
+        if not vals.get('membership_type'):
+            return super(ProductTemplate, self).write(vals)
+        for rec in self:
+            vals2 = vals.copy()
+            rec._correct_vals_membership_type(
+                vals2, vals.get('membership_type', rec.membership_type))
+            super(ProductTemplate, rec).write(vals2)
+        return True


pedrobaeza:
Although technically correct, this will affect the performance a lot. I would check if there's any `membership_type` and then perform the call to super independently. If not, call super for all of them together.

pedrobaeza (approved):

membership_variable_period/models/product_template.py
```diff
@@ -63,6 +63,10 @@ def create(self, vals):
...
     def write(self, vals):
-        self._correct_vals_membership_type(
-            vals, vals.get('membership_type', self.membership_type))
-        return super(ProductTemplate, self).write(vals)
+        if not vals.get('membership_type'):
+            return super(ProductTemplate, self).write(vals)
+        for rec in self:
+            rec._correct_vals_membership_type(
+                vals, vals.get('membership_type', rec.membership_type))
+            super(ProductTemplate, rec).write(vals)
```
pedrobaeza: Agh, another side effect here: please make a copy of vals for each:
```suggestion
            vals2 = vals.copy()
            rec._correct_vals_membership_type(
                vals2, vals.get('membership_type', rec.membership_type))
            super(ProductTemplate, rec).write(vals2)
```

pedrobaeza:
Let's fast-track it

/ocabot merge patch

pedrobaeza:
@chienandalu is this affecting 11.0 / 12.0 ?

chienandalu:
Looks like it does.

pedrobaeza:
OK, cherry-picked for v12 in 651f1d0f91ed86dac29038f7b8c6f58a56de5447 and v11 in b9b09dd503f977d103df1a94d77fc482a162c3df, and I see that @astirpe has included it as well in v13 migration.

PR #84, opened 2019-11-22
[13.0][MIG] membership_initial_fee

astirpe:

PR #99, opened 2020-12-14
[FIX] membership_initial_fee: associate members

chienandalu:
If the invoice partner is associated to the real member, we should
evalute past membership lines from that one

cc @Tecnativa TT27052

### DIFF
--- a/membership_initial_fee/models/account_invoice.py
+++ b/membership_initial_fee/models/account_invoice.py
@@ -47,10 +47,14 @@ class AccountInvoiceLine(models.Model):
                 product.initial_fee == 'none' or
                 self.invoice_id.type != "out_invoice"):
             return False
+        # If we are associated to another partner membership, evaluate that
+        # partner lines
+        partner = (
+            self.partner_id.associate_member or self.invoice_id.partner_id)
         # See if partner has any membership line to decide whether or not
         # to create the initial fee
         return not self.env['membership.membership_line'].search_count([
-            ('partner', '=', self.invoice_id.partner_id.id),
+            ('partner', '=', partner.id),
             ('account_invoice_line', 'not in', (self.id,)),
             ('state', 'not in', ['none', 'canceled']),
         ])


membership_initial_fee/models/account_invoice.py
```diff
@@ -47,10 +47,13 @@ def initial_fee_create_check(self, product=False):
                 product.initial_fee == 'none' or
                 self.invoice_id.type != "out_invoice"):
             return False
+        # If we are associated to another partner membership, evaluate that
+        # partner lines
+        partner = self.partner_id.associate_member or self.partner_id
```
pedrobaeza: ?
```suggestion
        partner = self.partner_id.associate_member or self.invoice_id.partner_id
```

chienandalu: There's a related field in the invoice line: https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_invoice.py#L1760-L1761

pedrobaeza: I don't have good experiences with this kind of fields being filled in the proper time, so if you don't mind...

chienandalu: Ok them :+1:

pedrobaeza (approved):

pedrobaeza (changes requested):
Wait, what about using:

https://github.com/OCA/vertical-association/blob/ac707fdc63c1f510d76db0a38357d43c3159f0aa/membership_delegated_partner/models/account_invoice.py#L52

?

pedrobaeza:
But having such method on base and then expanding it on the other module.

chienandalu:
I like the idea to avoid further issues, but the problem is that they both depend on the `membership` core module

pedrobaeza (approved):
OK, I see, it's a pity not to be able to do that.

pedrobaeza:
Fast-tracking for fixing the issue:

/ocabot merge patch

To be fw-port it to 13?

chienandalu:
I pinged @astirpe in his migration PR: https://github.com/OCA/vertical-association/pull/84

PR #100, opened 2020-12-31
[IMP] membership_withdrawal: allow to archive reasons

chienandalu:
cc @Tecnativa TT26867

### DIFF
--- a/membership_withdrawal/views/membership_withdrawal_reason_view.xml
+++ b/membership_withdrawal/views/membership_withdrawal_reason_view.xml
@@ -16,7 +16,18 @@
     <field name="arch" type="xml">
         <tree string="Membership withdrawal reasons" editable="top">
             <field name="name"/>
+            <field name="active" readonly="1" />
         </tree>
+    </field>
+</record>
+
+<record id="membership_withdrawal_reason_search_view" model="ir.ui.view">
+    <field name="model">membership.withdrawal_reason</field>
+    <field name="arch" type="xml">
+        <search>
+            <filter string="Active" domain="[('active','=', True)]" name="active_reasons"/>
+            <filter string="Archived" domain="[('active','=', False)]" name="archived_reasons"/>
+        </search>
     </field>
 </record>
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
And the UI?

chienandalu:
Well, when the field is present in the model you can always archive-unarchive selecting records in the tree view and going to *Action*

pedrobaeza:
But on v12 the usual thing is the smart-button, and I'm not sure buttons appear for this version.

chienandalu:
The thing is there's no proper form view defined for this model, only an editable tree

pedrobaeza:
OK, maybe include the field there?

chienandalu:
Mm, maybe a better improvement would be to add a search view for archived items
image

pedrobaeza (approved):

membership_withdrawal/views/membership_withdrawal_reason_view.xml
```diff
@@ -16,10 +16,22 @@
     <field name="arch" type="xml">
         <tree string="Membership withdrawal reasons" editable="top">
             <field name="name"/>
+            <field name="active" readonly="1" />
```
astirpe, line 19: I think this solution is ok, although I would prefer to have a form view defined for this model and a toggle button, to be consistent with the standard V12 archival feature.

membership_withdrawal/views/membership_withdrawal_reason_view.xml
```diff
@@ -16,10 +16,22 @@
...
     </field>
 </record>
 
+<record id="membership_withdrawal_reason_search_view" model="ir.ui.view">
+    <field name="model">membership.withdrawal_reason</field>
+    <field name="arch" type="xml">
+        <search>
+            <filter string="All" domain="['|', ('active','=', True), ('active','=', False)]" name="all_reasons"/>
+            <filter string="Active" domain="[('active','=', True)]" name="active_reasons"/>
+            <filter string="Archived" domain="[('active','=', False)]" name="archived_reasons"/>
```
astirpe, line 29: Only the "Archived" filter is needed, the other two filters are not strictly necessary and could be removed.

astirpe (approved):

membership_withdrawal/views/membership_withdrawal_reason_view.xml
pedrobaeza: Although the `All` filter can be removed, the other is useful for allowing to see both together (active and non active). They should be anyway split with `<separator>` elements for not combining with other filter elements.

pedrobaeza:
/ocabot merge minor

Please fw-port to v13/v14

PR #102, opened 2021-02-09
[FIX] membership_delegated/fee: Solves an incompatibility between both

chienandalu:
FW port from #54

Without changing inter-dependencies, using dynamic code detection.

TT27686 cc @Tecnativa


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this successive fw-ports:

/ocabot merge patch

marcelsavegnago (approved):

PR #105, opened 2021-03-26
[10.0][FIX] membership_extension: Fix refund validate if origin is void

rousseldenis:


rousseldenis:
@pedrobaeza @chienandalu

chienandalu (approved):
LGTM :+1:

rousseldenis:
> LGTM

Thanks.

IMHO something should be done to avoid searching on origin (a field that link together records or ...)

pedrobaeza:
A bit old for us, but straight-forward:

/ocabot merge patch

PR #108, opened 2021-06-29
[14.0][MIG] - membership_extension

sbejaoui:


membership_extension/models/membership_line.py
```diff
@@ -42,13 +42,17 @@ def _onchange_membership_date(self):
...
                 line.state = line.state or "none"
-            if line.account_invoice_id.invoice_payment_state == "paid":
-                line.state = "paid"
+            elif (
+                line.account_invoice_id.state == "posted"
+                and line.account_invoice_id.payment_state == "reversed"
+            ):
+                line.state = "canceled"
+            else:
+                super(MembershipLine, line)._compute_state()
```
chienandalu, line 55: Looking in the core method, I wonder if this override is needed anymore: https://github.com/odoo/odoo/blob/14.0/addons/membership/models/membership.py#L48-L78

membership_extension/tests/test_membership.py
```diff
@@ -363,19 +359,17 @@ def test_invoice(self):
...
         reversal = move_reversal.reverse_moves()
         refund = self.env["account.move"].browse(reversal["res_id"])
-        refund.journal_id.update_posted = True
-        refund.action_post()  # validate refund
         self.assertEqual("canceled", line.state)
         refund.button_cancel()
         self.assertEqual("paid", line.state)
         refund.button_draft()
-        self.assertEqual("paid", line.state)
+        self.assertEqual("invoiced", line.state)
```
chienandalu, line 372: Was it wrong before?

membership_extension/models/membership_line.py
sbejaoui: The case of invoice refunded is not covered in core method and it fixes the possible issue when `account_invoice_id`  is null

https://github.com/odoo/odoo/blob/ff6d0b16cd32a5ed56a2cc039f5226dc56f6db6a/addons/membership/models/membership.py#L48

membership_extension/tests/test_membership.py
sbejaoui: I am not sure how it was before, but now when the refund is canceled the original invoice is not automatically reconciled with the original payment.

sbejaoui:
Travis is red due to failing test in 14.0, I created an issue for it #109

membership_extension/tests/test_membership.py
rousseldenis: I think this is due to payment refactor.

rousseldenis:
@pedrobaeza @sbejaoui This one is green now.

rousseldenis (approved):
Code review

victoralmau (approved):
Code review OK.

Issue #109, opened 2021-06-29
Travis is red in 14.0


sbejaoui:

pedrobaeza:
It seems travis definition is not up to date. Maybe @sbidoul can update it.

rousseldenis:
> It seems travis definition is not up to date. Maybe @sbidoul can update it.

It's not travis related.

See this job just re launched on 14.0: https://travis-ci.com/github/OCA/vertical-association/jobs/511982017

I think there is a change in Odoo (but haven't looked furthermore)

pedrobaeza:
I was seeing incorrect builds (with exclamation) not getting to Odoo part:

https://travis-ci.com/github/OCA/vertical-association/jobs/520173116

rousseldenis:
@pedrobaeza I did a fix there for errors in 14.0 #110

pedrobaeza:
Thanks for the patch.

PR #110, opened 2021-06-30
[14.0][FIX] membership_delegated_partner: Use Form to test invoicing

rousseldenis:


rousseldenis:
Fixes #109 partly

membership_delegated_partner/tests/test_membership_delegate.py
```diff
@@ -97,23 +98,14 @@ def test_02_change_delegated_member(self):
...
-            }
-        )
+        move_form = Form(self.env['account.move'].with_context(default_move_type='out_invoice'))
+        move_form.partner_id = self.partner1
+        with move_form.invoice_line_ids.new() as line_form:
+            line_form.product_id = self.product
+            line_form.account_id = self.account
+            line_form.price_unit = 1.0
+        invoice = move_form.save()
+        invoice.write({"delegated_member_id": self.partner2.id})
```
pedrobaeza, line 110: Why not doing this on the Form?

pedrobaeza (approved):
Good for me if it's green

membership_delegated_partner/tests/test_membership_delegate.py
rousseldenis: I was waiting for that :smiley: Because the field is not in the Form, so not writeable there

rousseldenis: But IMHO, after this, tests should be enhanced to be done with Form as some values cannot be initialized.

sbejaoui (approved):

membership_delegated_partner/tests/test_membership_delegate.py
pedrobaeza: hehe, OK, but maybe it's because you are selecting the incorrect view? I see that the field is added in the proper one:

https://github.com/OCA/vertical-association/blob/14.0/membership_delegated_partner/views/account_move.xml

chienandalu (approved):

pedrobaeza:
I merge for greenifying the branch although my comment:

/ocabot merge patch

PR #111, opened 2021-09-14
[13.0][ADD] website_membership_random_order: Online Members Directory - Random order

pedrobaeza:
Online Members Directory - Random order
=======================================

This modules extends the online members directory for displaying the results in a random order, and displays a disclaimer about it at the top of the list.

Known issues / Roadmap
----------------------

- All the members will be displayed at once, without pager, as this is the only way to get a whole randomized list. On contrary, paging the content, on each render, we will get random elements belonging to that slice. Don't use this module with a long list of members.

@Tecnativa TT31670


website_membership_random_order/README.rst
```diff
@@ -0,0 +1,92 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Known issues / Roadmap
+======================
+
+- All the members will be displayed at once, without pager, as this is the only
+  way to get a whole randomized list. On contrary, paging the content, on each
```
chienandalu, line 40: Maybe you could use an already randomized value from the user session (i.e.: cookie id) and use it to get a reliable order. A snippet from: https://stackoverflow.com/a/19307329

```python
>>> import random
>>> x = [1, 2, 3, 4, 5, 6]
>>> random.Random(4).shuffle(x)
>>> x
[4, 6, 5, 1, 3, 2]
>>> x = [1, 2, 3, 4, 5, 6]
>>> random.Random(4).shuffle(x)
>>> x
[4, 6, 5, 1, 3, 2]
```

pedrobaeza: But it's not the intended goal, as for that session, the order will be always the same, and thus the top results won't change. We prefer to show everything and each time you access, see different top results.

chienandalu: Ok, ok. I guess it could be achieved with a more complex implementation, but for the present purposes it will suffice

chienandalu (approved):

website_membership_random_order/templates/website_membership_templates.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="index" inherit_id="website_membership.index">
+        <xpath expr="//t[@t-if='not memberships_partner_ids']" position="after">
+            <t t-else="">
+                <p
+                    style="text-align: center"
```
joao-p-marques: Could this be styled a little differently to help differentiate it from the rest of the page content?

joao-p-marques (approved):
LGTM :+1: Code a functional review.

Just a small suggestion, but nothing blocking

website_membership_random_order/templates/website_membership_templates.xml
pedrobaeza: Done:

Selección_018

joao-p-marques: Nice! Much better IMHO :smiley:

PR #118, opened 2022-01-25
[14.0] [IMP] membership_extension: show/hide fields on membership lines

primes2h:
membership_extension adds many features to membership base module. (see [README](https://github.com/OCA/vertical-association/blob/14.0/membership_extension/README.rst))

It also adds several fields to the membership list view on the member form.

- date_from (From)
- date_to (To)
- date_cancel (Cancel date)
- category_id (Membership category)

This PR allows to show/hide these fields on the view (defaults for new members: show date_from/date_to, hide date_cancel/category_id).

immagine

immagine

### DIFF
--- a/membership_extension/views/res_partner_view.xml
+++ b/membership_extension/views/res_partner_view.xml
@@ -40,20 +40,20 @@
                 expr="//field[@name='member_lines']/tree//field[@name='date']"
                 position="after"
             >
-                <field name="date_from" />
-                <field name="date_to" />
+                <field name="date_from" optional="show" />
+                <field name="date_to" optional="show" />
             </xpath>
             <xpath
                 expr="//field[@name='member_lines']/tree//field[@name='membership_id']"
                 position="after"
             >
-                <field name="category_id" />
+                <field name="category_id" optional="hide" />
             </xpath>
             <xpath
                 expr="//field[@name='member_lines']/tree//field[@name='date_to']"
                 position="after"
             >
-                <field name="date_cancel" />
+                <field name="date_cancel" optional="hide" />
             </xpath>
             <xpath expr="//field[@name='member_lines']/form" position="replace">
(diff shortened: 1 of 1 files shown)


membership_extension/views/res_partner_view.xml
```diff
@@ -40,20 +40,20 @@
                 expr="//field[@name='member_lines']/tree//field[@name='date']"
                 position="after"
             >
-                <field name="date_from" />
-                <field name="date_to" />
+                <field name="date_from" optional="hide" />
```
pedrobaeza: This should be shown by default IMO, and isn't this the most important info in the line?

primes2h: Ops, sorry. 
I swapped show/hide. :sweat_smile:

primes2h: Fixed.

pedrobaeza: But as said, do you think that 2 fields should be optional? Without them, the line loses its meaning.

primes2h: membership_extension adds many features, and not everyone needs all of them.
Besides, adding 4 fields in that view make it less readable as fields width gets smaller.
My idea is to show only the most important ones (date_from and date_to) in the view (missing in the original module).

What do you think?

pedrobaeza: Yeah, but as optional?

primes2h: They are shown by default but users could choose to hide them, more freedom.

If you think it's not worth it I can remove those options for date_from/date_to.

pedrobaeza (approved):
OK, let's keep it this way

mymage (approved):
LGTM

chienandalu (approved):

PR #123, opened 2022-06-23
[IMP] membership_extension

xavier-bouquiaux:
need a way to redefine what's an old membership
adding a method so we can override it in our project

### DIFF
--- a/membership_extension/models/res_partner.py
+++ b/membership_extension/models/res_partner.py
@@ -171,6 +171,15 @@ class ResPartner(models.Model):
                 partner.membership_stop = last_to
                 partner.membership_cancel = last_cancel
 
+    def _state_should_be_old(self, state):
+        self._membership_member_states()
+
+        return state == "none" and bool(
+            self.member_lines.filtered(
+                lambda r: r.state in self._membership_member_states()
+            )
+        )
+
     @api.depends(
         "free_member",
         "member_lines.state",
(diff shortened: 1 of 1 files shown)


membership_extension/models/res_partner.py
```diff
@@ -171,6 +171,14 @@ def _compute_membership_date(self):
                 partner.membership_stop = last_to
                 partner.membership_cancel = last_cancel
 
+    def _state_should_be_old(self, state):
+        member_states = self._membership_member_states()
+        if state == "none" and self.member_lines.filtered(
+            lambda r: r.state in member_states
+        ):
+            return True
+        return False
```
chienandalu:
```suggestion
        return state == "none" and bool(self.member_lines.filtered(
            lambda r: r.state in self._membership_member_states()
        ))
```

acsonefho (approved):

xavier-bouquiaux:
@chienandalu , i've applied your comment
could we merge this ?

membership_extension/models/res_partner.py
```diff
@@ -171,6 +171,15 @@ def _compute_membership_date(self):
                 partner.membership_stop = last_to
                 partner.membership_cancel = last_cancel
 
+    def _state_should_be_old(self, state):
+        self._membership_member_states()
+
```
chienandalu, line 176: This call is useless
```suggestion
```

PR #136, opened 2023-01-10
[FIX] membership_extension partner categories labels warning

PierrickBrun:
on the res.partner,  membership_categories and membership_category_ids have the same label. this can be confusing and creates a warning everywhere in the logs.

This PR changes the label of membership_categories to avoid that


PierrickBrun:
@legalsylvain FYI

chienandalu (approved):

PR #138, opened 2023-02-09
[16.0][IMP] membership_extension: check expiry on membership lines

PierrickBrun:
Before this PR, only membership state is changed when expiring.

This means a partner can have membership lines with status "paid" event if the date_to is in the past.

This PR fixes it, now membership lines are also updated when expiring


PierrickBrun:
Test failing fixed in #139

PierrickBrun:
rebased on #139

PR #139, opened 2023-02-21
[16.0][FIX] membership_extension: compute membership state: old member

PierrickBrun:
To compute if the partner is an old member we should check that any kind of membership line is associated with it, not valid ones.

Before:
Old members are considered Non member

After:
Old members status is applied correctly

### DIFF
--- a/membership_extension/models/res_partner.py
+++ b/membership_extension/models/res_partner.py
@@ -226,7 +226,7 @@ class ResPartner(models.Model):
                     if prior.get(line.state, 0) > prior.get(state):
                         state = line.state
                 if state == "none" and partner.member_lines.filtered(
-                    lambda r: r.state in member_states
+                    lambda r: r.state not in ["none"]
                 ):
                     state = "old"
                 partner.membership_state = state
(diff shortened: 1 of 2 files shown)


PierrickBrun:
I rebased manually

rafaelbn:
Hello @carmenbianca ! Do you agree? @fcvalgar ?

fcvalgar (commented):
After my functional review I detect:

The use case where the membership does not have a period the field "current membership status" is not calculated well.

What do you think @PierrickBrun ?


image

PierrickBrun:
> After my functional review I detect:
> 
> The use case where the membership does not have a period the field "current membership status" is not calculated well.
> 
> What do you think @PierrickBrun ?

I did not notice this, thank you. Indeed I think in that case the partner should have the "paid member" status

fcvalgar (approved):
LGTM, thank you @PierrickBrun

rafaelbn:
Hello @chienandalu @carmenbianca , do you wanna review?

chienandalu (commented):
My feeling here is that this method (and maybe others in the module) need a revamp and patches on top of it don't help the readibility of the thing.

The method is fully overriding the original method. This maybe had a meaning back in the day when the upstream logic was rather basic (https://github.com/odoo/odoo/blob/10.0/addons/membership/models/partner.py#L39-L48). But in late versions the logic has been evolving (https://github.com/odoo/odoo/commit/1f77dc369e2f318470872afa167fdd6e0f999874)

What do we need to extend here in the first place? If anything (otherwise the whole code could go away) can we extend the original method properly?

Also, why do we need to refefine all the attributes of the original field?

Issue #141, opened 2023-03-21
membership_extension:Group not applied by default to admin users


bvanjeelharia:
## Module
membership_extension

## Describe the bug
This module creates a new menu and applies on the Members App menu. But this group doesn't get applied to admin user(/root user) by default.It should be applied by default atleast to the admin user(and root user).

## To Reproduce
**Affected versions**:
All verisons
1.Install the membership_extension module
2.The Members app menu would not be visible to admin user (unless Manager group for membership module is applied to admin user)

**Expected behavior**
The group should be applied by default to admin(and root user)

bvanjeelharia:
I reverified in the module and found that the group is getting applied to admin user through demo file, so I believe it should be removed from there and added directly

xavivars:
This is still an issue. Should be `no stale` @bvanjeelharia

PR #142, opened 2023-03-21
[15.0] [FIX] membership_extension:Add memberhsip manager group to admin user

bvanjeelharia:
Fixes https://github.com/OCA/vertical-association/issues/141

### DIFF
--- a/membership_extension/__manifest__.py
+++ b/membership_extension/__manifest__.py
@@ -27,6 +27,5 @@
     "demo": [
         "demo/membership_category_demo.xml",
         "demo/product_template_demo.xml",
-        "demo/membership_security_demo.xml",
     ],
 }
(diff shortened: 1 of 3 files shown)


bvanjeelharia:
I reverified in the module and found that the group is getting applied to admin user through demo file, so I believe it should be removed from there and added directly.Please suggest if this makes sense or not

chienandalu (commented):
Thanks @bvanjeelharia 

Yes I think it's ok and that the demo declaration can go away

chienandalu (approved):
:+1:

pedrobaeza:
/ocabot merge minor

Please move it to upper version.

chienandalu:
Please forward port it to v16 if needed

xavivars:
v16 still has this problem

chienandalu:
> v16 still has this problem

Ok, @bvanjeelharia didn't make the fw. But any contribution is welcome :wink:

PR #143, opened 2023-03-22
[15.0][MIG] website_membership_random_order: Migration to version 15.0

stefan-tecnativa:
cc @Tecnativa TT42051

@pedrobaeza please review!


website_membership_random_order/controllers/main.py
```diff
@@ -0,0 +1,52 @@
...
+            "<int:country_id>",
+            "/members/association/<membership_id>/country/<int:country_id>",
+            "/members/association/<membership_id>/country/<country_name>-"
+            "<int:country_id>/page/<int:page>",
+            "/members/association/<membership_id>/country/<int:country_id>/"
+            "page/<int:page>",
+        ],
+        type="http",
+        auth="public",
+        website=True,
```
chienandalu: This is missing the `sitemap` argument (https://github.com/odoo/odoo/blob/15.0/addons/website_membership/controllers/main.py#L32)

Anyway @pedrobaeza is there a reason to not do a simpler overriding with

```python
@route()
def members(...)
```

chienandalu (changes requested):
It isn't working for me. The members always appear in the same order. Does it work for you?

stefan-tecnativa:
> It isn't working for me. The members always appear in the same order. Does it work for you?

I've observed that in both 13.0 and 14.0 runbots, the behavios that you are describing happens also for the members that has their state on "free member". But for those who has their state on "paid" this is not happening and they are displayed randomly. 
Searching I've found this, but I'm not sure if it has any relation with our problem: https://github.com/odoo/odoo/blob/15.0/addons/website_membership/controllers/main.py#L44

chienandalu:
True, they're randomized inside their membership category and free members don't get randomized at all. We can live with that IMO

website_membership_random_order/models/membership_line.py
```diff
@@ -16,6 +16,7 @@ def search(self, args, offset=0, limit=None, order=None, count=False):
         won't be able to show all the records due to the pager and each time returning
         random results.
         """
+
```
pedrobaeza: Why adding this empty line?

stefan-tecnativa: I was debugging and I forgot to remove the line when I removed the debugger method. Changed.

chienandalu (approved):
:+1:

website_membership_random_order/controllers/main.py
pedrobaeza: I didn't know that such option is available, so if everything is kept, then better to do it that way.

chienandalu: Here's the reference: 
- https://www.odoo.com/documentation/16.0/developer/reference/backend/http.html?highlight=controller#controllers
- https://www.odoo.com/documentation/16.0/developer/reference/backend/http.html?highlight=route#odoo.http.route

pedrobaeza: Great, then please do, @stefan-tecnativa

stefan-tecnativa: Changes done. Thanks for the explanations.

pedrobaeza:
With the latest change, `from odoo import http` should go away.

pedrobaeza (approved):
/ocabot merge nobump

PR #146, opened 2023-04-04
[FIX] 14.0 membership_extension cron activate members (via date_from)

PierrickBrun:


rafaelbn:
Could you please explain Step to reproduce (STR) ? Thank you!

PierrickBrun:
Before:

If you create a membership line that starts in the future, the partner's membership status is never checked and never becomes a member

After:

One the start date of the membership is reached, the partner becomes a member

PR #147, opened 2023-04-04
[FIX] 16.0 membership_extension cron activate members (via date_from)

PierrickBrun:


rafaelbn:
Could you please explain Step to reproduce (STR) ? Thank you!

PierrickBrun:
> Could you please explain Step to reproduce (STR) ? Thank you!

Before:

If a membership is created in the future (date_from > now) then the partner will not be a member, this is OK
But when the date changes and date_from become <= now, then nothing happens and the partner does not become a member: NOK

After:

Both cases are OK because the cron checks if there are members to enable

PierrickBrun:
I rebased on #139

PR #149, opened 2023-05-16
[12.0][FIX] membership_delegated_partner Weblate it.po error

mymage:
Update of it.po that generate an error on Weblate

### DIFF
--- a/membership_delegated_partner/i18n/it.po
+++ b/membership_delegated_partner/i18n/it.po
@@ -36,9 +36,8 @@ msgstr "Fattura"
 
 #. module: membership_delegated_partner
 #: model:ir.model,name:membership_delegated_partner.model_account_invoice_line
-#, fuzzy
 msgid "Invoice Line"
-msgstr "Fattura"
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

pedrobaeza:
Why are you doing this here instead of Weblate?

https://odoo-community.org/page/translate

mymage:
> Why are you doing this here instead of Weblate?
> 
> https://odoo-community.org/page/translate

Because when I try to translate it on Weblate, it raise a server error.
Let me know if there is another way to do/fix it.

pedrobaeza:
I think that has been something temporary today. Please try again.

mymage:
> I think that has been something temporary today. Please try again.

Ok, no problem. I opened this because in another case ([management-system/#499](https://github.com/OCA/management-system/pull/499)) I did this way.
I will try for some days and comment here on positive/negative result.

rafaelbn (approved):
Similar to https://github.com/OCA/management-system/pull/501

PR #153, opened 2023-09-25
[16.0][MIG] membership_delegated_partner: Migration to version 16.0

carolinafernandez-tecnativa:
- Standard migration to v16

@Tecnativa

TT45234

@pedrobaeza @chienandalu


membership_delegated_partner/__manifest__.py
```diff
@@ -1,16 +1,17 @@
...
-    "version": "15.0.1.0.0",
+    "version": "16.0.1.0.0",
     "development_status": "Production/Stable",
     "category": "Membership",
     "author": "Tecnativa, Onestein, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/vertical-association",
     "license": "AGPL-3",
     "summary": "Delegate membership on a specific partner",
-    "depends": ["membership"],
+    "depends": ["membership", "sale"],
```
chienandalu: How so?

carolinafernandez-tecnativa: > How so?

Not neccesary, I got sale module installed so when i was running test for this module I got a null constraint.

chienandalu (approved):
Code an functional review :+1: Thanks!

membership_delegated_partner/tests/test_membership_delegate.py
```diff
@@ -0,0 +1,146 @@
...
+        """Delegates membership to partner 2"""
+        invoice = self.env["account.move"].create(
+            {
+                "name": "Test Customer Invoice",
+                "move_type": "out_invoice",
+                "partner_id": self.partner1.id,  # Invoicing partner
+                "delegated_member_id": self.partner2.id,  # Delegate membership to
+            }
+        )
+        self.env["account.move.line"].create(
```
pedrobaeza: Instead of this (and doing the change from `invoice_line_ids[0]` to `line_ids[0]`, you should use the one2many field `invoice_line_ids` and create everything in one shot on the previous `account.move`>`create` call. The same for the other part where things are done the same way.

pedrobaeza:
I think you should add the type of line as `product`.

carolinafernandez-tecnativa:
> I think you should add the type of line as `product`.

Thnks! Done :)

pedrobaeza (approved):
/ocabot merge nobump

PR #157, opened 2023-10-05
[16.0][IMP] membership_extension & _variable_period: Make sure necessary fields are always truthy

carmenbianca:
This is an additional sanity check to make sure that fields that are expected to be available are actually available.

The situation here is a little tricky. Especially as pertains date_from and date_to; these fields are `required=True` nowhere. Not on the product, not on the membership invoice, and not on the membership line. However, they _are_ set as required in the product and membership line views. If these values are ever empty, it's a sure sign that something went wrong somewhere, because the user should not be able to reach such a state via the UI.

One sure way to reach that state was via the demo data, however, which had no start and end dates. This is now fixed.

The situation for the variable period fields is similar. These fields should never be empty when the membership type is variable, and it's not typically possible to empty these values via the UI. For convenience's sake, I simply set them to required an sich, with a default value.

### DIFF
--- a/contract_membership_delegated_partner/tests/test_delegated_partner_contract.py
+++ b/contract_membership_delegated_partner/tests/test_delegated_partner_contract.py
@@ -9,7 +9,13 @@ class TestMembershipDelegateSetup(TestContractBase):
     def setUpClass(cls):
         super(TestMembershipDelegateSetup, cls).setUpClass()
         cls.partner2 = cls.env["res.partner"].create({"name": "Mrs. Odoo"})
-        cls.product_1.membership = True
+        cls.product_1.write(
+            {
+                "membership_date_from": "2023-01-01",
+                "membership_date_to": "2023-01-02",
+                "membership": True,
+            }
+        )
         cls.contract.delegated_member_id = cls.partner2
 
     def test_01_generate_and_delegate(self):
--- a/membership_extension/models/product_template.py
+++ b/membership_extension/models/product_template.py
@@ -4,7 +4,8 @@
 
 from datetime import timedelta
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ProductTemplate(models.Model):
@@ -32,3 +33,13 @@ class ProductTemplate(models.Model):
             if record.company_id and record.membership_category_id.company_id:
                 if record.membership_category_id.company_id != record.company_id:
                     record.membership_category_id = False
+
+    @api.constrains("membership_date_from", "membership_date_to", "membership")
+    def _check_membership_dates(self):
+        if self.filtered(
+            lambda record: record.membership
+            and (not record.membership_date_from or not record.membership_date_to)
+        ):
+            raise ValidationError(
+                _("A membership product must have a start date and an end date.")
+            )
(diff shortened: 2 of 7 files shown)


membership_extension/models/product_template.py
```diff
@@ -32,3 +33,14 @@ def _compute_membership_category_id(self):
             if record.company_id and record.membership_category_id.company_id:
                 if record.membership_category_id.company_id != record.company_id:
                     record.membership_category_id = False
+
+    @api.constrains("membership_date_from", "membership_date_to", "membership")
+    def _check_membership_dates(self):
+        for record in self:
+            if record.membership:
+                if not record.membership_date_from or not record.membership_date_to:
```
chienandalu:
```suggestion
        if self.filtered(lambda x: x.membership and (not x.membership_date_from or not x.membership_date_to):
```

membership_variable_period/models/product_template.py
```diff
@@ -79,3 +88,15 @@ def write(self, vals):
...
+        "membership_date_from",
+        "membership_date_to",
+        "membership",
+        "membership_type",
+    )
+    def _check_membership_dates(self):
+        for record in self:
+            if record.membership:
+                if record.membership_type == "fixed":
+                    return super(ProductTemplate, record)._check_membership_dates()
```
chienandalu: You're breaking your loop here. Maybe what you want to do is:

```suggestion
        return super(ProductTemplate, self.filtered(lambda x: x.membership_type == "fixed"))._check_membership_dates()
```

chienandalu (commented):
Hi @carmenbianca I'm ok with your suggestions. Just a couple of comments:

membership_variable_period/models/product_template.py
carmenbianca: Good catch!

carmenbianca:
@chienandalu I've implemented and pushed your suggestions. Thank you for the review.

chienandalu (approved):
Thanks :)

contract_membership_delegated_partner/tests/test_delegated_partner_contract.py
```diff
@@ -9,7 +9,13 @@ class TestMembershipDelegateSetup(TestContractBase):
     def setUpClass(cls):
         super(TestMembershipDelegateSetup, cls).setUpClass()
         cls.partner2 = cls.env["res.partner"].create({"name": "Mrs. Odoo"})
-        cls.product_1.membership = True
+        cls.product_1.write(
+            {
+                "membership_date_from": "2023-01-01",
```
pedrobaeza, line 14: Why do you need to change this?

carmenbianca: @pedrobaeza Because the new constraint won't let a membership product have no start/end dates.

```python
    @api.constrains("membership_date_from", "membership_date_to", "membership")
    def _check_membership_dates(self):
        if self.filtered(
            lambda record: record.membership
            and (not record.membership_date_from or not record.membership_date_to)
        ):
            raise ValidationError(
                _("A membership product must have a start date and an end date.")
            )
```

(Interestingly, the write function is sensitive to the order of key-value pairs. If `"membership": True` appears before the other items, the constraint complains. Not sure why that happens.)

pedrobaeza: It's weird, as in other places, you are removing such fields, but it's not a big deal.

/ocabot merge patch

PR #159, opened 2023-10-11
[FIX] membership_extension: No return on super call

hhgabelgaard:
With the return we only got the state computed for the first line in self

### DIFF
--- a/membership_extension/models/membership_line.py
+++ b/membership_extension/models/membership_line.py
@@ -43,6 +43,8 @@ class MembershipLine(models.Model):
                     self.date_to = date_to
 
     def _compute_state(self):
+        # Compute methods should not return
+        # pylint: disable=missing-return
         for line in self:
             if isinstance(line.id, models.NewId) or not line.account_invoice_id:
                 line.state = line.state or "none"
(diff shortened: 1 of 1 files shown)


membership_extension/models/membership_line.py
```diff
@@ -43,6 +43,8 @@ def _onchange_membership_date(self):
                     self.date_to = date_to
 
     def _compute_state(self):
+        # Compute methods should not return
+        # pylint: disable=missing-return
```
pedrobaeza, line 47: Instead of this, just `return True` at the end.

membership_extension/models/membership_line.py
```diff
@@ -52,7 +54,7 @@ def _compute_state(self):
             ):
                 line.state = "canceled"
             else:
-                return super(MembershipLine, line)._compute_state()
+                super(MembershipLine, line)._compute_state()
```
chienandalu, line 57: A more performant way would be to avoid splitting the super line by line. Something like:
```python
no_invoice_lines = self.filtered(lambda line: isinstance(line.id, models.NewId) or not line.account_invoice_id)
cancelled_lines = self.filtered(lambda line: line.account_invoice_id.state == "posted" and line.account_invoice_id.payment_state == "reversed")
for line in no_invoice_lines:
    line.state = line.state or "none"
cancelled_lines.state = "cancelled"
return super(MembershipLine, (self - no_invoice_lines - cancelled_lines)._compute_state()
```

pedrobaeza:
Superseded by #163

PR #161, opened 2023-11-20
[16.0][REF/ADD] Create membership_type, membership_custom_date

carmenbianca:
`membership_custom_date` is a custom module used by Akretion/UGESS. It used to depend directly on `membership_variable_period`, which functionally made no sense, but was necessary because we re-used the `membership_type` field on product.template.

I've factored out that field into `membership_type` and made both `membership_variable_period` and `membership_custom_date` depend on that new module.

(Almost) no history was lost for `membership_custom_date` in copying the module over. 99% of the code was already in a single commit.


membership_type/models/product_template.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @api.constrains(
+        "membership",
+        "membership_type",
+    )
+    def _check_membership_dates(self):
+        return super(
+            ProductTemplate,
+            self.filtered(lambda record: record.membership_type == "fixed"),
+        )._check_membership_dates()
```
carmenbianca: The constraints fields don't seem right to me.

carmenbianca: fixed

membership_variable_period/views/product_template_views.xml
```diff
@@ -9,8 +9,7 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="membership.membership_products_form" />
```
nayatec: should be `membership_type.membership_products_form` which actually defines `membership_type` no ?

nayatec (changes requested):
A part from the comment i made, LGTM

membership_variable_period/views/product_template_views.xml
carmenbianca: You are right! Corrected.

nayatec (approved):

PR #164, opened 2024-03-18
[15.0][MIG]membership_prorate:Migration to 15.0

bizzappdev:


membership_prorate/models/account_invoice.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        date_from = product.membership_date_from
+        date_to = product.membership_date_to
+        return date_from, date_to
+
+    def _prepare_invoice_line_prorate_vals(self, invoice_line):
+        product = invoice_line.product_id
+        date_invoice = invoice_line.move_id.invoice_date or fields.Date.today()
+        date_from, date_to = self._get_membership_interval(product, date_invoice)
+        if date_invoice < date_from:
```
edlopen, line 29: I've stumbled upon an issue when configuring a membership product with the prorate field set to true and an the membership type set to "Variable periods". As "date_to" is not set it raises an error while comparing a date and a boolean.

Maybe the functions _get_membership_interval should consider this case and calculate the dates.

edlopen (changes requested):
Found an error when date_to is not set. Could you reviw this please?

membership_prorate/models/account_invoice.py
bizzappdev: @edlopen, basically, for your exact scenario special module is there [membership_prorate_variable_period](https://github.com/OCA/vertical-association/tree/14.0/membership_prorate_variable_period). which is a bridge module (auto install True) for membership_prorate and membership_variable_period modules. which is solving the problem 
so when the membership_prorate_variable_period module is migrated, it will solve your issue.

edlopen (approved):

rafaelbn (approved):
/ocabot merge patch

PR #166, opened 2024-04-15
[ADD] membership_pricelist_assignment

PlantBasedStudio:


PlantBasedStudio:
I'm currently having trouble passing a unit test. It sends the "old" state to a subscription and checks whether the subscription has been set to "old". This is the case, but the test obviously considers that it doesn't.

PlantBasedStudio:
@OCA/core-maintainers
I would like to have a review on this module

membership_pricelist_assignment/data/membership_data.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+    Copyright 2024 Moka Tourisme (https://www.mokatourisme.fr/).
+    @author Damien Horvat <ultrarushgame@gmail.com>
+    @author Romain Duciel <romain@mokatourisme.fr>
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo>
```
petrus-v, line 8: it seems to be a common practice to keep ir.cron records as noupdate to let administrator control crons

```suggestion
<odoo noupdate="1">
```

membership_pricelist_assignment/data/membership_data.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field name="model_id" ref="membership.model_membership_membership_line" />
+        <field name="state">code</field>
+        <field name="code">model._cron_update_membership_lines_pricelist()</field>
+        <field name="interval_type">days</field>
+        <field name="interval_number">1</field>
+        <field name="numbercall">-1</field>
+        <field
+            name="nextcall"
+            eval="(DateTime.now() + timedelta(days=1)).strftime('%Y-%m-%d 00:00:00')"
+        />
```
petrus-v, line 20: I think this is not necessary

```suggestion
```

membership_pricelist_assignment/i18n/fr.po
```diff
@@ -0,0 +1,120 @@
+# Translation of Odoo Server.
```
petrus-v, line 1: Not sure if OCA bot will do it automatically  a `membership_pricelist_assignment.pot` file should be versioned as well ?

membership_pricelist_assignment/models/membership.py
```diff
@@ -0,0 +1,168 @@
...
+from odoo import api, fields, models
+
+
+class MembershipLineInherit(models.Model):
+    """Extend the membership membership line model."""
+
+    _inherit = "membership.membership_line"
+
+    pricelist_id = fields.Many2one("product.pricelist", string="Pricelist", store=True)
+    user_id = fields.Many2one("res.users", string="Users", store=True)
```
petrus-v, line 16: are you realy need this field in this module, I can't see what the purpose ?

membership_pricelist_assignment/models/product.py
```diff
@@ -0,0 +1,55 @@
...
+    Added a product_hierarchy field to store the products that will not be considered.
+    """
+
+    _inherit = "product.template"
+
+    pricelist_id = fields.Many2one(
+        "product.pricelist",
+        string="Associated Pricelist",
+        help="Pricelist to add to the member",
+        store=True,
```
petrus-v, line 21: This is useless fields are stored by default unless you add a compute method
```suggestion
```

petrus-v (commented):
Thanks for contributing.

I'm a bit confused, this code does things that I wouldn't expected according the module description and seems complicated for its announced goal.

Can you split different concerns in different modules ?

If I understand properly, you try to define a price-list on partners and apply it at some point (at last while generating the invoice) ?

If so could you remove all the code which is not useful on that.

It's quite a long time I've played with the membership module,  but I've some doubt according choices that have been made when and how to apply the price list, can we speak about it on Monday?

PR #168, opened 2024-05-02
16.0 mig membership prorate variable period

edlopen:
Standard migration to 16.0

@moduon MT-1833

@yajo @EmilioPascual @Shide @rafaelbn review this PR when you can. Thank you!


yajo (approved):

yajo:
@chienandalu wanna review this one?

EmilioPascual (approved):

edlopen:
Can you check it functionally @fcvalgar? Thanks!

fcvalgar (commented):
I think the calculation of the prorate is not correct because it does not include the current day and the final day. In the memberships both days are usually included.

image

image

image

edlopen:
Hi @fcvalgar , I've reviewed the two day difference that you commented. It is kind of a bummer because it seems that is missing the first and the last day and it lead us to misunderstood of what the problem really is.

I'll try to not enter too much in the details but this is a matter of decimal precision in the quantities.

In your case you are prorrating for the last 236 days in the year. If the quantity of a year is 1, our prorrated quantity should be 0.646575(...) years but in the invoice line will store only 0.64 which is that value rounded to two decimals. That lead us to $24 difference. So a membership that cost $1 a day make us think that we are missing the starting and ending date but, sadly, not.

To sum up, the difference that you notice is a precission error in the prorrate calculation. Because this is a migration I would consider that this PR is ok to merge but the roadmap will be updated to cover this issue later on.

I have a couple of ideas of how make this run propelly but I feel that is out of the scope of this PR.

Also, what do you think @EmilioPascual ?

fcvalgar (approved):
Functional review,

LGTM, thank you @edlopen

yajo:
I think it's worth it to record your discoveries in the module's roadmap/known issues.

EmilioPascual:
Perhaps error is in the rounding in this line https://github.com/OCA/vertical-association/blob/16.0/membership_prorate/models/account_move_line.py#L37, but it's in another module. 
Anyway migration is ok.

edlopen:
@rafaelbn this PR is ok to merge. We've all agreed on this migration to move on. Thank you!

rafaelbn (approved):
/ocabot migration patch

rafaelbn:
@EmilioPascual @edlopen about this:

> Perhaps error is in the rounding in this line https://github.com/OCA/vertical-association/blob/16.0/membership_prorate/models/account_move_line.py#L37, but it's in another module.
Anyway migration is ok.

Please open an issue in githug and a subtas in our side. 

thank you! 😄 You are awesome!

PR #174, opened 2024-09-16
[16.0] [ADD] membership_activity

ByteMeAsap:
This module allows to track activities of members.This is a base module that allows to define some basic membership activities. Other modules are to be added on top of this module for tracking github,gitlab related activities

PR #176, opened 2024-09-18
[16.0] [ADD] gitlab

ByteMeAsap:
This module does not provide a functionality by itself but a base model
to implement gitlab related features


chienandalu (commented):
Wrong repo?

flotho:
think the same thing @chienandalu

ByteMeAsap:
> Wrong repo?

@chienandalu , can you suggest appropriate repo for adding this module? I would be using this module as a dependent module for creating a module for logging gitlab commits,PRs as membership activities(refer https://github.com/OCA/vertical-association/pull/174)

chienandalu:
I guess server-backend would be more appropiate. What do you think @pedrobaeza

pedrobaeza:
I new repository may be created, like we have interface-github. Anyway, I think this module is very weak and should incorporate more things to be useful for anyone, apart from #174, which is not very related with Gitlab itself.

ByteMeAsap:
> I new repository may be created, like we have interface-github. Anyway, I think this module is very weak and should incorporate more things to be useful for anyone, apart from #174, which is not very related with Gitlab itself.

@pedrobaeza So, should this module be kept as a part of this repo? or should we incorporate the features of this module in the other module for membership_activity?

pedrobaeza:
I don't still understand the goal of the `gitlab` module, and as said, it's very weak for a complete module, as it's not tied at all to Gitlab in any sense.

tarteo:
@pedrobaeza This module is a technical module it's not meant for the user to do anything. It's meant for other modules to use that need a gitlab connection. https://github.com/OCA/vertical-association/pull/174 is not related to gitlab, but we have created another module that logs / creates membership activity based on commits, pull requests (merge requests), opening issues etc etc. which we will also propose.

But yes this module is not supposed to be here I would suggest server-backend.

ByteMeAsap:
Created a PR in server-backend - https://github.com/OCA/server-backend/pull/311

PR #180, opened 2024-10-17
[15.0][FIX] membership_extension: post() is deprecated since v14: Replace with action_post()

hhgabelgaard:
post() is deprecated since v14: https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L2681

So we rarely get the state updated on the membership line. Especially on creditnotes

This should probably be ported to 14.0 and future versions

### DIFF
--- a/membership_extension/models/account_move.py
+++ b/membership_extension/models/account_move.py
@@ -38,9 +38,9 @@ class AccountMove(models.Model):
                 lines.write({"date_cancel": False})
         return res
 
-    def post(self):
+    def action_post(self):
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Fast-tracking this obvious one:

/ocabot merge patch

pedrobaeza:
Please next time follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

PR #163, opened 2024-01-08
[15.0][FIX] membership_extension: Don't return early

pedrobaeza:
Follow-up of #159 after not getting answer

Don't return before consuming whole recordset.

@Tecnativa

PR #181, opened 2024-10-17
[16.0][FIX] membership_extension: Don't return early

pedrobaeza:
Forward-port of #163 

Don't return before consuming whole recordset.

@Tecnativa


pedrobaeza:
Fast-tracking the forward-port

/ocabot merge patch

PR #182, opened 2024-10-17
[17.0][FIX] membership_extension: Don't return early

pedrobaeza:
Forward-port of #163 and #181

Don't return before consuming whole recordset.

@Tecnativa


pedrobaeza:
Let's finish the fw-port chain:

/ocabot merge patch

PR #184, opened 2024-10-21
[17.0][FIX] membership_extension: post() is deprecated since v14

hhgabelgaard:
Forward port of #180 and #183

Replaced by action_post(): https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L2681 So we rarely get the state updated on the membership line. Especially on creditnotes


pedrobaeza (approved):
/ocabot merge patch

PR #187, opened 2024-11-15
[17.0][MIG] website_membership_random_order: Migration to 17.0

victoralmau:
Migration to 17.0

@Tecnativa TT51560


chienandalu:
Testing on runbot, I set serveral free members. They show up always in the same order:

image

victoralmau:
The random mode does not work with free members, only with `membership.membership_line` records. 
This behavior is the same as in previous versions.

chienandalu:
Shouldn't it be in the known issues section?

pedrobaeza (approved):
OK to add the mention about free members.

chienandalu (approved):
:+1:

PR #191, opened 2024-12-18
[FIX] membership_extension: support restarted memberships

yajo:
When a membership got cancelled and, later, restarted, we still had a cancellation date set, leading to the sensation that it was still cancelled.

With this refactor, when a cancellation date is earlier than a future start date, cancellation info is removed from the partner.

Before:
image

After:
image

BTW I improved a few things:
- Removed old dead code that is no longer necessary in v16.
- Fix a test that was checking what happened when a membership line was cancelled before its start date. It seems a wrong use case because memberships are only supposed to be cancelled during the membership, and not before starting it.
- Free members need to be considered members too.
- Migration script.

@moduon MT-8237


fcvalgar (approved):
Functional review, 

LGTM, thank you @yajo.

image

rafaelbn (approved):
/ocabot merge patch

PR #192, opened 2025-01-24
[16.0][IMP] membership_extension: Add manager group to default user template

tarteo:


pedrobaeza:
The same as in spreadsheet: a new user shouldn't have this by default IMO.

Issue #190, opened 2024-12-17
[16.0][membership_extension] 'membership_last_start' does not correctly reflect the start of the latest membership period after a break


fcvalgar:
## Module
membership_extension

## Describe the bug
It has been observed that the membership_last_start field does not accurately calculate the start date of the most recent membership period when there has been a break in membership. Instead of taking the start date of the new period following the gap, the field appears to rely on the last active fee, resulting in inaccurate information about the continuity of the membership.

This issue makes it difficult to accurately analyze when a member has rejoined and can be considered "new" in a given period. The correct behavior should clearly identify the start date of the latest continuous membership period after any interruption, rather than carrying over outdated information.

References to Previous PRs:

Previous issue: [#126](https://github.com/OCA/vertical-association/issues/126)
Migration to Odoo 16: [#132](https://github.com/OCA/vertical-association/pull/132)

## To Reproduce

1. Create an initial membership (e.g., from 01/01/2020 to 31/12/2020).
2. Leave a gap with no membership activity (e.g., 2021-2022 without any membership).
3. Register a new membership starting from 17/12/2023 to 31/12/2023.
4. Register another membership from 01/01/2024 to 31/12/2024 (a full year).
5. Check the membership_last_start field. Currently, it does not show the correct start date for the latest period, as it carries over previous information.

**Expected behavior**
membership_last_start should be set to the actual start date of the latest continuous membership period following any break. In the example provided, it should properly reflect the late 2023 and 2024 memberships, offering a clear view of the member as “new” or returning during that period, thereby facilitating accurate analysis of membership base evolution.

**Additional context**
A fix to the membership_last_start calculation would help provide more reliable metrics, especially after the migration to Odoo 16 discussed in PR [#132](https://github.com/OCA/vertical-association/pull/132). Any suggestions or contributions from the community to resolve this issue are welcome.

@moduon MT-8237 @yajo 

FYI @carmenbianca @chienandalu

rafaelbn:
Let fix it @fcvalgar @yajo , I agree 😄 with ❤️

yajo:
I agree there is a problem here.

There are 3 involved fields: https://github.com/OCA/vertical-association/blob/f19fdbcba2d3210f133f73bdd1b8cf5d4e6f5d32/membership_extension/models/res_partner.py#L38-L61

According to my tests, `membership_stop` mostly behaves as expected in any case. However, the other 2 are not always well understood.

[...]

Issue #126, opened 2022-07-29
membership_last_start is not correct


PRMA974:
## Module
Membership extension

## Describe the bug
Le renouvellement d'une adhésion ne donne pas la bonne information sur le début de la dernière adhésion (membership_last_start)

## To Reproduce
**Affected versions**: 12, 13, 14

1. Créer une adhésion de date à date : ie du 01/01/2021 au 31/12/2021
2.Créer un renouvellement d'adhésion : ie du 01/01/2022 au 31/12/2022

**Expected behavior**
Voici les informations attendues
Date de début d'adhésion : 01/01/2021
Dernière date d'adhésion : 01/01/2022
Date de fin d'adhésion : 31/12/2022

**Additional context**
Or, il y a une erreur car membership_last_start = membership_start soit 01/01/2021 au lieu de 01/01/2022.
Je pense qu'il y a une erreur dans la boucle avec les variables last_from et date_from dans le fichier res.partner.py.

PRMA974:
Je constate également qu'en opérant un tri par date sur les adhésions et en remettant l'adhésion à un état null puis à un état "membre payant" cela résoud le problème...peut-être un début de solution à appliquer dans le code ?

carmenbianca:
Reporting the same bug on version ~~15~~ 14. Copied from internal bug tracker:

> Scenario:
>
> Partner has three memberships:
>
> 1. Membership A going from 2022-08-01 to 2022-08-31.
> 2. Membership B going from 2022-09-01 to 2022-09-30.
> 3. Membership C going from 2022-10-01 to 2022-10-31.
>
> The current date, if it matters, is 2022-09-09 (during Membership B).
>
> What should the value of `membership_last_start` be? In my patch, it's 2022-10-01. On the previous behaviour, it's 2022-08-01.

The patch:

```patch
commit 798bb24970f67fe2fe5109efebf3d96355cc8782
Author: Carmen Bianca Bakker <carmen@coopiteasy.be>
Date:   Wed Sep 7 16:28:51 2022 +0200

    [FIX] membership_extension: Use latest start date for Membership Last Start Date
    
    I'm not sure what the behaviour prior to this patch is.
    
    Signed-off-by: Carmen Bianca Bakker <carmen@coopiteasy.be>

diff --git a/odoo/temp-oca-src/membership_extension/models/res_partner.py b/odoo/temp-oca-src/membership_extension/models/res_partner.py
index 9cc4757..1882d61 100644
[...]

PR #193, opened 2025-01-27
[FIX] membership_extension: recover good old behavior of membership last start

yajo:
93c8e75450efc4e08b211e7e4e4a93be438fee28 was supposed to fix https://github.com/OCA/vertical-association/issues/126 but in reality the problem was the misunderstanding of the feature.

Here I'm reverting that commit and changing the help attribute of the fields to be more clear when explaining what each one means. The feature is now tested.

The details are explained in https://github.com/OCA/vertical-association/issues/190#issuecomment-2551150567 (option A).

Fixes https://github.com/OCA/vertical-association/issues/190 
@moduon MT-8237
cc @fcvalgar


fcvalgar (approved):
Functional review, 

LGTM, thank you @yajo.

image

EmilioPascual (approved):

rafaelbn (approved):
👍🏼 Fine!

Ready to merge! ❤️

PR #204, opened 2025-03-05
[16.0][FIX] membership_delegated_partner + membership_extension + membership_variable_period + membership_prorate: Post-install test + fallback to load CoA

victoralmau:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with the error:

`odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale`

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/membership_extension/tests/test_membership.py
+++ b/membership_extension/tests/test_membership.py
@@ -10,16 +10,25 @@ from psycopg2 import IntegrityError
 
 from odoo import fields
 from odoo.exceptions import UserError, ValidationError
-from odoo.tests import common
+from odoo.tests import common, tagged
 from odoo.tools import mute_logger
 
 
-@freeze_time("2025-01-01")
+@tagged("-at_install", "post_install")
 class TestMembership(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        date_today = fields.Date.today()
+        if not cls.env.company.chart_template_id:
+            # Load a CoA if there's none in current company
+            coa = cls.env.ref("l10n_generic_coa.configurable_chart_template", False)
+            if not coa:
+                # Load the first available CoA
+                coa = cls.env["account.chart.template"].search(
+                    [("visible", "=", True)], limit=1
+                )
+            coa.try_loading(company=cls.env.company, install_demo=False)
+        date_today = fields.Date.from_string("2025-01-01")
 
         cls.account_bank = cls.env["account.account"].create(
             {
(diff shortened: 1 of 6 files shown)


membership_extension/tests/test_membership.py
```diff
@@ -10,16 +10,25 @@
 
 from odoo import fields
 from odoo.exceptions import UserError, ValidationError
-from odoo.tests import common
+from odoo.tests import common, tagged
 from odoo.tools import mute_logger
 
 
-@freeze_time("2025-01-01")
+@tagged("-at_install", "post_install")
```
yajo, line 17: Why unfreezing? 🤔

membership_prorate_variable_period/models/product_template.py
```diff
@@ -4,14 +4,20 @@
...
     _inherit = "product.template"
 
     def _get_next_date(self, date, qty=1):
         next_date = super(ProductTemplate, self)._get_next_date(date)
+        test_condition = not config["test_enable"] or (
+            config["test_enable"]
+            and self.env.context.get("test_membership_prorate_variable_period")
+        )
+        if not test_condition:
+            return next_date
```
yajo: suggestion: make variable purpose clearer by changing its name and definition.

```suggestion
        testing_other_modules = config["test_enable"] and not self.env.context.get("test_membership_prorate_variable_period")
        if testing_other_modules:
            return next_date
```

membership_extension/tests/test_membership.py
victoralmau: This is necessary to avoid side effects:
- If another module inherits from this one the tests will be blocked.
- If the test is in post_install, all the tests that are executed will use that date.

Some references:
- https://github.com/OCA/credit-control/pull/434#issuecomment-2690137011
- https://github.com/OCA/field-service/pull/1340#issuecomment-2711247458

pedrobaeza: It's not being unfreezed, but moving the freezing at method level due to that side effects.

pedrobaeza (approved):
/ocabot merge patch

PR #206, opened 2025-04-21
[17.0][IMP] website_membership_gamification: Avoid duplicated label

pedrobaeza:
```
Two fields (gamification_badge_ids, badge_ids) of res.users() have the same label: Badges. [Modules: None and gamification]
```

@Tecnativa


pedrobaeza:
Ouch, didn't bump the module version, sorry.

PR #151, opened 2023-07-04
[NEW] membership_delegated_partner_line

gdgellatly:
Draft of new module to support delegation at the line level on invoices.

PR #207, opened 2025-04-28
[16.0][IMP] membership_delegated_partner: allow multiple delegated members per invoice

flaenen:
It is now possible to define a delegate member per invoice line (in addition to the one defined on the invoice). A customer can therefore be invoiced for the memberships of several partners.

A delegated member defined on an invoice line has priority over the delegated member defined globally on the invoice.

### DIFF
--- a/membership_delegated_partner/README.rst
+++ b/membership_delegated_partner/README.rst
@@ -45,8 +45,12 @@ permission on.
 Usage
 =====
 
-#. In an invoice with membership lines, choose a delegated partner.
+#. In an invoice with membership lines, choose a delegated partner directly on
+   the lines or globally on the invoice.
 #. The membership line will go to the delegated partner.
+
+A delegated partner defined on an invoice line has priority over the delegated
+partner defined globally on the invoice.
 
 Bug Tracker
 ===========
--- a/membership_delegated_partner/i18n/fr.po
+++ b/membership_delegated_partner/i18n/fr.po
@@ -2,26 +2,28 @@
 # This file contains the translation of the following modules:
 # * membership_delegated_partner
 #
-# Translators:
-# OCA Transbot <transbot@odoo-community.org>, 2017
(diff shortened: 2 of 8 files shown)


chienandalu (commented):
Did you see https://github.com/OCA/vertical-association/pull/151 ? There were some issues spotted there with this approach. (fyi @gdgellatly )


In any case, I'm missing test coverage here

flaenen:
> Did you see #151 ? There were some issues spotted there with this approach. (fyi @gdgellatly )
> 
> In any case, I'm missing test coverage here

I'll take a look at this PR, thank you.

gdgellatly:
So in #151 I have passed the torch on that and I have not been involved in a long time, can't quite remember who to, but they are doing the work. 151 has actually been in production for years now for the OCA and IMO should eventually supersede membership_delegated_partner as the preferred module. 

The reason for a separate module was technical at the time, something to do with computed fields in the upstream module, I don't really remember. That issue may no longer exist in newer versions.

Given the OCA's reliance on the feature, I imagine this is something that the relevant working group will want to be involved in.

huguesdk (approved):
i think this approach is much simpler than #151 which introduces a separate and incompatible module.

lgtm.

gdgellatly:
> i think this approach is much simpler than #151 which introduces a separate and incompatible module.
> 
> lgtm.

Sure, recent tendencies for people to just do what they want and not worry about anyone who depends on the existing behaviour, but at the time you would never dream of disrupting an existing module and its existing dependencies. That is why it was seperate after discussion with maintainer

chienandalu (commented):
Reviewing you approach I agree with @gdgellatly that it would be compatible with both use cases. And this would not introduce incompatibility with the existing base, I think.

In any case, the extra test cases for membership from lines are very much needed, IMO. Maybe you could borrow them from #151

PR #214, opened 2025-10-30
[MIG] [18.0] membership delegated partner line

gfcapalbo:
Migration from v14 to v18.

module requires  uninstalling membership_delegated_partner  and to run a sql script before that to be installed.

we have no mechanism for  "blocks" or "incompatible_with" in manifest for now
I would like to add init_hook blocking installation if membership_delegated_partner. the error will return the basic instructions to follow to correctly install the module.
This small improvement, unfortunately makes test fail, asking for OCA-way to bypass small problem.

Delivering the Current MR has same functionality as old v14. 
If we want the pre_init_hook (ready to go on another branch) i need to understand how could this be compatible with OCA CI.


membership_delegated_partner_line/models/membership_line.py
```diff
@@ -0,0 +1,61 @@
...
+
+    partner = fields.Many2one(compute="_compute_partner", store=True, readonly=False)
+
+    @api.depends(
+        "account_invoice_line.delegated_member_id",
+        "account_invoice_line.move_id.partner_id",
+    )
+    def _compute_partner(self):
+        """Change associated membership lines if delegated member is changed.
+        NOTE: This is required due to the weird way that account.move in parent
```
Kiplangatdan, line 19: @gfcapalbo Perhaps this was already solved in [v18](https://github.com/OCA/vertical-association/blob/18.0/membership_delegated_partner/models/membership_line.py#L17) and this override is not needed anymore?

membership_delegated_partner_line/views/account_move.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='invoice_line_ids']/list/field[@name='product_id']"
+                position="after"
+            >
+                <field
+                    name="delegated_member_id"
+                    attrs="{'column_invisible':[('parent.is_membership_invoice', '!=', True)], 'invisible':[('is_membership', '!=', True)]}"
```
Kiplangatdan: @gfcapalbo `attrs` is no longer used.

Kiplangatdan (changes requested):

membership_delegated_partner_line/models/membership_line.py
gfcapalbo: Yes. but the two modules are incompatible.  membership_delegated_partner_line does not depend on membership_delegated_partner,   in fact they are incompatible and should not be installed together.

So repeating the  _compute_partner here as is done in   https://github.com/OCA/vertical-association/blob/18.0/membership_delegated_partner/models/membership_line.py#L17        is necessary.

membership_delegated_partner_line/views/account_move.xml
gfcapalbo: reformatted in v18 normal attributes

Kiplangatdan (approved):

membership_delegated_partner_line/models/account_move.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    is_membership_invoice = fields.Boolean(compute="_compute_is_membership_invoice")
+
+    @api.depends("invoice_line_ids", "line_ids", "move_type", "line_ids.product_id")
+    @api.onchange("invoice_line_ids", "line_ids", "move_type")
```
hbrunn, line 13: why do we need the onchange?

membership_delegated_partner_line/models/account_move.py
```diff
@@ -0,0 +1,21 @@
...
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    is_membership_invoice = fields.Boolean(compute="_compute_is_membership_invoice")
+
+    @api.depends("invoice_line_ids", "line_ids", "move_type", "line_ids.product_id")
```
hbrunn, line 12:
```suggestion
    @api.depends("invoice_line_ids", "line_ids", "move_type", "line_ids.product_id.membership")
```

membership_delegated_partner_line/models/membership_line.py
hbrunn: I suggest to also express this in the module's manifest file's `excludes` key

hbrunn (changes requested):
commit history is missing

PR #215, opened 2026-02-02
[18.0][FW] membership_extension: multiple ports from 16.0

yajo:
Port from 16.0 to 18.0:
- #191
- #193

@moduon MT-13535


fcvalgar (approved):
Functional review.

LGTM, thank you @yajo.

Image

membership_extension/models/membership_line.py
```diff
@@ -19,9 +19,7 @@ class MembershipLine(models.Model):
     )
     date_from = fields.Date(readonly=False)
     date_to = fields.Date(readonly=False)
-    state = fields.Selection(
-        compute="_compute_state", inverse="_inverse_state", store=True, readonly=False
-    )
+    state = fields.Selection(compute="_compute_state", readonly=False)
```
chienandalu: Why converting it to non storable?

yajo: This happens just as a fwport of https://github.com/OCA/vertical-association/pull/191. I can't remember the details, but it is part of fixing an issue.

chienandalu: But this is the kind of change that could cause issues to others using the module... The rationale isn't clear from the original commit either...

yajo: I removed it because it appeared to be unnecessary code. Looking at the implementation, `_inverse_state` was empty (just returning True), and `_compute_state` had no dependencies defined, which suggested the code wasn't functioning as intended.

Since this change has been reviewed, tested, and running successfully in production on v16 for about a year now, I see no reason to block its migration, honestly.

chienandalu: > and _compute_state had no dependencies defined, which suggested the code wasn't functioning as intended.

You have to look at the core, where those depends are defined. Could be the error to not keep the @api.depends() decorator?

https://github.com/odoo/odoo/blob/eed2bf97536cc9dc2fb471936e2422e2313e3009/addons/membership/models/membership.py#L36-L48

Sorry, but to me removing an exisisting column from the core looks kind of potentially problematic...

chienandalu: My bad. The override wasn't removing any column as the original method already defines the compute and the store flag and that original setting remains if not changed.

chienandalu (approved):
/ocabot merge minor

PR #219, opened 2026-03-25
[16.0][FIX] membership_extension: remove access right

mihien:
Remove a line granting all accesses to products to every users


mihien:
Hi @chienandalu , I want to remove an access rule that was added in #53 , do you remember what was the context of that rule? Is it needed ?

yajo (approved):

chienandalu (approved):
wow! 😅

Issue #216, opened 2026-03-02
[DISCUSSION] Future of "membership" module, removed of Odoo CE since v18.3


arnaudlayec:
Hello all,

I am writing this issue to discuss and define a proposal to manage the fact that Odoo S.A. simply removed the "**membership**" module from Odoo Community Edition since v18.3, in favor of a "**partnership**" module which does not render the same functionalities.

### Context

You may verify that in browsing Odoo v19.0 "addons" folder: https://github.com/odoo/odoo/tree/19.0/addons, the "membership" module does not exist anymore. "Membership" is not even in Enterprise edition, it has just been removed in 
Also see this PR and commit:
* https://github.com/odoo/odoo/pull/197228
* https://github.com/odoo/odoo/commit/787ee30aba7110b59cd76ba4ef1864332e306108

I'm also opening this issue in the context of work ordered by and realized for the OCA, which wants to improve its membership process in their recently-migrated v18.0 Odoo (call for quotation of September 2025, RFQ2). In the context of this work, we decided to rely on the **membership** module still available in v18.0, but which will not exist anymore in this format at next OCA's Odoo upgrade.

### Proposal

I'm wondering about the role of the community and scenarios we can think of to move forward.
Without have a lot of background, I'm pointing out a few facts in my knowledge and possibilities for the OCA.
I think the PSC of **vertical-association** must be consulted.

Facts:
* "**membership**" module was old, and I understand not easily inheritable or fixable
* it **is** used by Odoo customers, not only the OCA (I know for a fact at least 1 big engineer school in Switzerland)
* the new module "**partnership**", presented as a replacement, is rather a module to manage corporate partners like https://www.odoo.com/partners

Possibilities (very simply):
1. write to Odoo S.A. to clarify the situation
2. fork "**membership**" in v18.2 and start maintaining a "**membership_oca**" starting v19.0 in this repository: could be the opportunity to "clean" the module, and offer an alternative to customers previously using in in v18.2 and below
3. anything else?

For OCA RFQ: @bguillot @TumbaoJu 
For vertical-association editors & PSC: @chienandalu @pedrobaeza @rafaelbn
For OCA's internal tools working group: @petrus-v  @thomaspaulb

pedrobaeza:
[...]

PR #224, opened 2026-06-25
[19.0][ADD+MIG] membership,membership_extension: Migration to 19.0 + membership_extension merged into membership

victoralmau:
Related to https://github.com/OCA/vertical-association/issues/216

Migration to 19.0

Please @pedrobaeza can you review it?

@Tecnativa TT63407


chienandalu (commented):
Maybe we could simplify downstream implementation adding some hooks. For example, the partner_id computation in https://github.com/OCA/vertical-association/pull/225

membership/views/product_template_views.xml
```diff
@@ -0,0 +1,196 @@
...
+                <field name="list_price" string="Membership Fee" />
+                <field name="categ_id" column_invisible="True" />
+                <field name="uom_id" column_invisible="True" />
+                <field name="type" column_invisible="True" />
+            </list>
+        </field>
+    </record>
+    <record id="membership_products_kanban" model="ir.ui.view">
+        <field name="name">product.template.kanban</field>
+        <field name="model">product.template</field>
```
eduezerouali-tecnativa, line 61:
```suggestion
        <field name="model">product.template</field>
        <field name="priority">50</field>
```

eduezerouali-tecnativa (changes requested):
Tested on runboat. LGTM + code review.

membership/views/product_template_views.xml
victoralmau: Why is this priority necessary?

eduezerouali-tecnativa: Because product.template's kanban view is a very generic, widely-inherited view. Without an explicit priority, this inheritance uses the default value (16), same as most other extensions of this view, so the order in which it gets applied depends purely on module load order.
That's non-deterministic and can cause this inheritance to be overridden by, or to override, unrelated modules' xpath changes on the same view. Setting priority="50" makes it apply after the default-priority inheritances, avoiding conflicts with other modules extending the same base view.
imagen

carlos-lopez-tecnativa: Done. Please update your review.

carlos-lopez-tecnativa (approved):
LGTM

XtremXpert:
Functional review from a real 19.0 production deployment.

We run this migration in production at **Les Maisons d'à côté Lanaudière**, a
Québec nonprofit operating two adapted residences. Odoo 19 Community, no
Enterprise modules. Figures below are read from the live database today:

| | |
|---|---|
| Partners with a membership state | **32** (29 `paid`, 2 `invoiced`, 1 `old`) |
| `membership.membership_line` records | **58** |
| Membership products | 4 (individual / partner member, 2025 and 2026) |
| Period covered | 2025-01-01 → 2026-12-31 |
| Invoices generated and posted | **58** |
| Dues invoiced | **900.00 CAD** |

Membership state transitions, dues invoicing and the yearly renewal across two
periods all behave correctly. No functional issue found.

## On the `priority` discussion

@eduezerouali-tecnativa's request is correct and worth applying, but I think the
stated mechanism is what's blocking agreement here — so, concretely:

**None of the four `product.template` views in `membership/views/product_views.xml`
are inheritances.** `membership_product_search_form_view`, `membership_products_tree`,
`membership_products_kanban` and `membership_products_form` are all **standalone
views**, referenced explicitly by `view_id` in the membership products action. So
this isn't about inheritance ordering.

The real reason is view *selection*: three of those four already carry
`priority=50`; `membership_products_kanban` is the only one without it, so it
falls back to the default **16** — the same priority as core's
`product.product_template_kanban_view`. That's a genuine tie in the default-view
lookup for `product.template`.

To be fair and not overstate it: **this is not an observable bug.** `membership`
depends on `product`, so `product` always installs first and gets the lower id.
Measured on our production database:

```
priority=16  id=869   product.product_template_kanban_view      ← selected
priority=16  id=3114  membership.membership_products_kanban
_get_view('kanban') → id=869
```

Core wins. But it wins *by id order*, i.e. correct by accident. Since the view is
named explicitly by its action, it never needs to win the default lookup at all —
`priority=50` simply states that intent and makes the kanban consistent with its
three siblings. It looks like a plain oversight rather than a decision.

We run `membership` in production with `priority=50` applied to that view; no
regression.

Otherwise this LGTM, and I'd be glad to approve once the suggestion lands.

---

*Reviewed with the assistance of an AI agent (Claude); all figures above were read
from our production database and verified by me before posting.*

eduezerouali-tecnativa (approved):

pilarvargas-tecnativa (approved):

pedrobaeza:
@sbidoul please another one to add to PyPi

sbidoul:
Reserved on PyPI. Given the number of interested people on the related issue, I'd give them more than a week to review. Maybe also notify maintainers of membership_extension.

pedrobaeza:
The PR is here since one month ago, and we are the maintainers of `membership_extension` (in fact, we are doing this for Spanish Association AEOdoo).

/ocabot merge nobump

membership/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+        "views/product_template_views.xml",
+        "views/res_partner_views.xml",
+        "views/membership_category_view.xml",
+        "report/report_membership_views.xml",
+    ],
+    "demo": [
+        "demo/membership_category_demo.xml",
+        "demo/product_template_demo.xml",
+    ],
+    "maintainers": ["rafaelbn", "yajo"],
```
sbidoul, line 27: Still it's strange that these are declared as maintainers and have not visibly been involved on the PR.

sbidoul:
> The PR is here since one month ago

https://github.com/OCA/vertical-association/issues/216#issuecomment-4983137473 is only one week old, though...

membership/__manifest__.py
pedrobaeza: Uhm, they should add them later, as we were maintaining these modules since 8.0... anyway, as always, we are very responsive for any change needed. They can just comment.

membership/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+{
+    "name": "Members",
+    "version": "19.0.1.1.0",
+    "category": "Membership",
+    "development_status": "Mature",
+    "author": "Odoo, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/vertical-association",
+    "depends": ["account"],
```
MiquelRForgeFlow, line 11:
```suggestion
    "depends": ["account", "partnership"],
```
why not this?

pedrobaeza: Because we don't want to depend on such module that is not bringing value.

PR #226, opened 2026-06-26
[19.0][MIG] website_membership: Migration to 19.0

victoralmau:
Migration to 19.0

Locked by: 
- [x] `membership`: https://github.com/OCA/vertical-association/pull/224

@Tecnativa TT63423

PR #227, opened 2026-07-01
[19.0][MIG] website_membership_gamification: Migration to 19.0

victoralmau:
Migration to 19.0

Locked by: 
- [x] `website_membership`: https://github.com/OCA/vertical-association/pull/226

Please @pilarvargas-tecnativa can you review it?

@Tecnativa TT63424


website_membership_gamification/models/res_partner.py
```diff
@@ -0,0 +1,66 @@
...
+                ]
+                sel.commercial_partner_id.gamification_badge_ids = badges
+
+    def _compute_has_related_users(self):
+        for sel in self:
+            sel.has_related_users = any(a.active for a in sel.user_ids)
+
+    def action_grant_badge_wizard(self):
+        user_hr_gamification_module = self.env["ir.module.module"].search(
+            [("name", "=", "user_hr_gamification")]
```
juancarlosonate-tecnativa: I don't see that `user_hr_gamification` module existing anywhere

juancarlosonate-tecnativa (changes requested):

website_membership_gamification/wizard/gamification_badge_user_wizard.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2022 Manuel Regidor <manuel.regidor@sygel.es>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="partner_view_badge_wizard_reward" model="ir.ui.view">
```
pilarvargas-tecnativa, line 5: Odoo already has its own wizard, the difference being that it allows you to select the user, but perhaps it could be simplified by using the core wizard and passing the user in the action itself. What do you think?

https://github.com/odoo/odoo/blob/06a46b083fde5688d71c5602553316c9e7160b6a/addons/gamification/wizard/grant_badge.xml#L2

Image

pedrobaeza: ping @victoralmau

victoralmau: I've reused a large part of the existing wizard in `gamification`.

pilarvargas-tecnativa (approved):

website_membership_gamification/views/res_partner_views.xml
```diff
@@ -0,0 +1,44 @@
...
+                    string="Grant a Badge"
+                    type="object"
+                    class="btn btn-secondary mb-4"
+                    name="action_grant_badge_wizard"
+                />
+                    <field
+                    name="gamification_badge_ids"
+                    mode="kanban"
+                    context="{'kanban_view_ref': 'website_membership_gamification.website_membership_gamification_badge_user_kanban_view'}"
+                /> />
```
juancarlosonate-tecnativa: typo?

juancarlosonate-tecnativa (commented):
If I have two child contacts with the same parent and I grant a badge to one of them, the parent gets that same badge. But if I then grant a badge to the other child, the parent's badge gets replaced. Shouldn't the parent contact have both badges?

victoralmau:
> If I have two child contacts with the same parent and I grant a badge to one of them, the parent gets that same badge. But if I then grant a badge to the other child, the parent's badge gets replaced. Shouldn't the parent contact have both badges?

I don't fully understand the use case you're describing, but that behavior https://github.com/OCA/vertical-association/blob/f8bcf398cc420b54f64bb44d275fd4b4da757f2d/website_membership_gamification/models/res_partner.py#L28 has been that way since version 15.0.

juancarlosonate-tecnativa (approved):
> I don't fully understand the use case you're describing, but that behavior
> 
> https://github.com/OCA/vertical-association/blob/f8bcf398cc420b54f64bb44d275fd4b4da757f2d/website_membership_gamification/models/res_partner.py#L28
> 
> has been that way since version 15.0.

I know it wasn't introduced in this migration. I was just asking if that was intentional, since it didn't seem that way to me. But okay, LGTM then.

pedrobaeza:
I think it depends on the linked users. Anyway, let's move for now.

/ocabot merge nobump

